PR #1, opened 2023-05-22
add bootz proto

marcushines:


proto/bootz.proto
```diff
@@ -0,0 +1,167 @@
...
+message ControlCard {
+  string part_number = 1;
+  string serial_number = 2;
+  int32 slot = 3;
+}
+
+// The status of each control card (i.e. initialized or not). For first install
+// of the chassis, both control cards are in NOT_INSTALLED state. For 
+// install a hot-swapped standby, the status of the active card is INSTALLED, 
+// and that of the standby is NOT_INSTALLED. For FF chassis, the serial number
```
dplore, line 79:
```suggestion
// and that of the standby is NOT_INSTALLED. For fixed form-factor chassis, the serial number
```

proto/bootz.proto
```diff
@@ -0,0 +1,167 @@
...
+// of the card on the chassis.
+message ControlCard {
+  string part_number = 1;
+  string serial_number = 2;
+  int32 slot = 3;
+}
+
+// The status of each control card (i.e. initialized or not). For first install
+// of the chassis, both control cards are in NOT_INSTALLED state. For 
+// install a hot-swapped standby, the status of the active card is INSTALLED,
```
dplore, line 78: nit: Is ACTIVE a better word than INSTALLED?

proto/bootz.proto
```diff
@@ -0,0 +1,167 @@
...
+    // progress.
+    string server_trust_cert = 4;
+    // Boot configuration is specified as structured data.
+    BootConfig boot_config = 5;
+    Credentials credentials = 6;
+    gnsi.pathz.v1.UploadRequest pathz = 7;
+    gnsi.authz.v1.UploadRequest authz = 8;
+    gnsi.certz.v1.UploadRequest certificates = 9;
+  } 
+  repeated BootstrapDataResponse responses = 1;
```
dplore, line 112: What is the scenario where there is more than one response when GetBootstrapDataRequest only contains one serial number?  Is this to allow multiple server_trust_cert and/or multiple SoftwareImage (for multiple url's where it could be downloaded from)?

proto/bootz.proto
```diff
@@ -0,0 +1,167 @@
...
+}
+
+message Credentials {
+  repeated gnsi.credentialz.v1.AuthorizedKeysRequest credentials = 1;
+  repeated gnsi.credentialz.v1.AuthorizedUsersRequest users = 2;
+  repeated gnsi.credentialz.v1.PasswordRequest passwords = 3;
+}
+
+message BootConfig {
+  string hostname = 1;
```
dplore, line 138: wouldn't hostname just be part of the *_config"?  Or maybe there is a use case for hostname to be outside of the config?

proto/bootz.proto
```diff
@@ -0,0 +1,167 @@
...
+  // Proprietary key-value parameters that are required as part of boot
+  // configuration (e.g., feature flags, or vendor-specific hardware knobs).
+  google.protobuf.Struct metadata = 4;
+  // Native format vendor configuration.
+  bytes vendor_config = 5;
+  // JSON rendered OC configuration.
+  bytes oc_config = 6;
+  // Bootloader key-value parameters that are required as part of boot
```
dplore, line 146: is this also a proprietary set of key values?  Does it make sense to just make part of metadata?

proto/bootz.proto
```diff
@@ -0,0 +1,167 @@
...
+  // configuration.
+  google.protobuf.Struct bootloader_config = 7;
+
+}
+
+// The device reports the status of applying BootstrapData using this RPC.
+// The status_message is a human-readable message indicating the nature of 
+// failure, if any.
+message ReportStatusRequest {
```
dplore, line 155: Should there be an IN_PROGRESS indicator and periodic update from the client to the boot strap server??  OS install can take a long time

dplore (commented):
A few comments

proto/bootz.proto
marcushines: i think that conflates different things
this is answering is the card installed - it doesn't imply which card is active/standby or any thing about the operational state of the card

proto/bootz.proto
marcushines: the "active" sup registers both itself and any other sup's in the chassis

proto/bootz.proto
marcushines: they are not the same data elements

proto/bootz.proto
dplore: I think the text difficult to understand because 'INSTALLED' is not clearly defined.  Does install mean the physical card is plugged in?  Or does it mean an operating system is installed?  Typically devices (outside of the whitebox ecosystem)  with factory defaults do include an operating system which is "installed".  

After reading this a few times, it seems like installed might be referring to the state of the bootstrap process.  ie: an indication of whether the bootstrap process has completed successfully or not.  

I think a clarification is needed

proto/bootz.proto
```diff
@@ -0,0 +1,167 @@
...
+}
+
+enum ControlCardStatus {
+  CONTROL_CARD_STATUS_UNSPECIFIED = 0;
+  CONTROL_CARD_STATUS_NOT_INSTALLED = 1;
+  CONTROL_CARD_STATUS_INSTALLED = 2;
+}
+
+message GetBootstrapDataResponse {
+  message BootstrapDataResponse {
```
rnhaddad, line 94: Should we add the nonce in here? The nonce needs to be part of the signed message

README.md
```diff
@@ -1,2 +1,262 @@
...
+## Detailed Design
+
+### Boot Procedure
+
+#### **API flow**
+
+1. DHCP Discovery of Bootstrap Server
+    1. Device sends DHCP messages, containing the mac-address of the active control card. The DHCP server has been configured with all possible mac-addresses of the device, and responds with the static IP address of the bootstrap server.
+    2. DHCP server also assigns an IP address and a gateway to the device.
+    3. The DHCP response option code should be 70 (Picked from the set of reserved/private use option codes).
```
rnhaddad, line 198: Sztp work flow uses option 136/143 in dhcp response message to inform redirect server information to client. Any reason why not use these?

proto/bootz.proto
marcushines: where do you think it shouldbe added?

rnhaddad: As last field within  message BootstrapDataResponse e.g.:
string nonce = 1001;

rnhaddad: But ideally, i would create a new message to avoid having it in the repeated filed:

message BootstrapDataResponseSigned {
     repeated BootstrapDataResponse responses = 1;
     string nonce = 1001;
}

In this case we solve 2 problems:
1. the nonce is set only once.
2. The signed message is clear and confined (Easily computed from bytes)

README.md
marcushines: what is 136? i see 143 that looks right

rnhaddad: From the RFC text https://datatracker.ietf.org/doc/rfc8572/ 

8.2.  DHCPv6 SZTP Redirect Option

   The DHCPv6 SZTP Redirect Option is used to provision the client with
   one or more URIs for bootstrap servers that can be contacted to
   attempt further configuration.

      0                   1                   2                   3
      0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
     +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
     |       option-code (136)       |          option-length        |
     +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
     .           bootstrap-server-list (variable length)             .
     +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+

     * option-code: OPTION_V6_SZTP_REDIRECT (136)

proto/bootz.proto
vijaysgr: // The status of each control card (i.e. initialized or not). For first install
// of the chassis, both control cards are in NOT_INSTALLED state. For 
// install a hot-swapped standby, the status of the active card is INSTALLED,
// and that of the standby is NOT_INSTALLED. For fixed form-factor chassis,
// the serial number is that of the chassis, and status is NOT_INSTALLED.


Suggested reword of the entire comment (lines 77-81).
// The status of each control card (bootstrapped or not). At the 
// beginning of the bootstrap process (i.e. before a ReportStatus
// has completed with BOOTSTRAP_STATUS_SUCCESS), both 
// control cards are in NOT_INSTALLED state. 
// When a control-card is hot-swapped, the bootz process is
// performed with the swapped control-card having a status of
// NOT_INSTALLED, and the other card in INSTALLED state.

dplore: Building off your suggestion below.  I also changed the enum to use NOT_INITIALIZED and INITIALIZED to be consistent with the readme which uses these terms.

If an `INITIALIZED` control card is removed and reinserted, should it still be in `INITIALIZED` state?  Or should it be `NOT_INITIALIZED`?  I would expect it to remain `INITALIZED`, yes?

// ControlCardState indicates if the bootstrap process has 
// successfully completed.
//
// At the beginning of the bootstrap process (i.e. before a 
// ReportStatus has completed with BOOTSTRAP_STATUS_SUCCESS),
// all control cards (1 or 2) are in NOT_INITIALIZED state.
// If bootstrap successfully completes for a control-card, the 
// ControlCardStatus changes to INITIALIZED.
//
// Once a control-card is in INITIALIZED state, it may remain in that
// state indefinitely, even if removed and reinstalled in a chassis.  
// 
// When there are 2 control-cards present and INITIALIZED and 1 is 
// removed, the remaining control card remains in INITIALIZED state.
// If a new control-card is inserted which has never completed the
// bootstrap process, it will start with NOT_INITIALIZED state.
message ControlCardState {
  string serial_number = 1;
  ControlCardStatus status = 2;
}

enum ControlCardStatus {
  CONTROL_CARD_STATUS_UNSPECIFIED = 0;       // the bootstrap process status is not reported
  CONTROL_CARD_STATUS_ NOT_INITIALIZED = 1;  //  the bootstrap process has not successfully completed
  CONTROL_CARD_STATUS_ INITIALIZED = 2;            //  the bootstrap process has successfully completed
}

dplore (changes requested):
Provided suggestion on ControlCardState

dplore (commented):
LGTM

PR #6, opened 2023-07-12
Hines

marcushines:


client/client
robshakir, line 1: Remove binary -- suggest to add something to `.gitignore` that avoids this in the future.

proto/bootz.proto
```diff
@@ -103,42 +103,43 @@ message ControlCardState {
     CONTROL_CARD_STATUS_INITIALIZED = 2; 
   }
 
+  // Serial must align with the serial numberpre
   string serial_number = 1;
   ControlCardStatus status = 2;
 }
 
-message GetBootstrapDataResponse {
+message BootstrapDataResponse {
```
robshakir, line 111: What's our expectation on stability of the API at this point? I presume we're OK with making backwards incompatible changes since this is still early?

server/entitymanager/entitymanager.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"github.com/openconfig/bootz/proto/bootz"
+	"github.com/openconfig/bootz/server/service"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+type InMemoryEntityManager struct{}
+
```
robshakir, line 12: Can we add docstrings for each of these exported methods? I think probably they would be best where the `interface` is defined and then you can just mark them as needed to implement the interface here?

server/service/service.go
```diff
@@ -0,0 +1,95 @@
+package service
```
robshakir, line 1: Package comment missing.

server/service/service.go
```diff
@@ -0,0 +1,95 @@
...
+import (
+	"context"
+
+	"github.com/openconfig/bootz/proto/bootz"
+	"github.com/openconfig/gnmi/errlist"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+type EntityLookup struct {
```
robshakir, line 12: What is the definition of an entity here? (I think this might be a gap in my understanding of bootz.) Are we saying that an entity is a device, or it's just something that can run bootz?

server/service/service.go
```diff
@@ -0,0 +1,95 @@
...
+	Manufacturer string
+	SerialNumber string
+}
+
+type ChassisEntity struct {
+	BootMode string
```
robshakir, line 18: It looks like you use this as an enumerated type -- I'd suggest to define a type to avoid magic string matching.

testdata/README.md
robshakir, line 1: Wrap to 80 characters ideally please.

testdata/generate.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+	vendorCAPrivPem, err := pemEncode(x509.MarshalPKCS1PrivateKey(vendorCAPriv), "RSA PRIVATE KEY")
+	if err != nil {
+		log.Exit(err)
+	}
+	if err := writeFile(vendorCAPrivPem, "vendorca_priv.pem"); err != nil {
+		log.Exit(err)
+	}
+
+	//Generate PDC.
```
robshakir, line 198:
```suggestion
	// Generate PDC.
```

robshakir (changes requested):
Neat to have something to test against here -- bunch of comments in-line, mostly Go ones, I didn't spot anything that seemed out of whack within the bootz logic.

client/README.md
```diff
@@ -0,0 +1,15 @@
+# Bootz Client Emulator
+
+The code located in this directory is intended to emulator a typical Bootz client. Where appropriate, some device-specific functions such as downloading and image or applying a config are mocked out and are simply logged.
```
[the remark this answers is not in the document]
marcushines: shouldn't they be aligned?

client/client.go
```diff
@@ -0,0 +1,308 @@
...
+				PartNumber:   "123B",
+			},
+		},
+	}
+
+	log.Infof("%v chassis %v starting with SecureOnly = %v", chassis.Manufacturer, chassis.SerialNumber, !*insecureBoot)
+
+	// 1. DHCP Discovery of Bootstrap Server
+	// This step emulates the retrieval of the bootz server IP
+	// address from a DHCP server. In this case we always connect to localhost.
```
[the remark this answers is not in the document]
marcushines: i kinda like the wording as it aligns with the bootz process - the text clearly states what we do differently here

proto/bootz.proto
```diff
@@ -103,42 +103,43 @@ message ControlCardState {
     CONTROL_CARD_STATUS_INITIALIZED = 2; 
   }
 
+  // Serial must align with the serial numberpre
```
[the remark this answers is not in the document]
marcushines: i think this was merge error

proto/bootz.proto
marcushines: for now

server/entitymanager/entitymanager.go
marcushines: not sure why the lint didn't complain on this

server/server.go
```diff
@@ -0,0 +1,45 @@
+// Bootz server reference implementation.
```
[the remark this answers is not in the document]
marcushines: i don't follow?

server/service/service.go
marcushines: control card or chassis

server/service/service.go
```diff
@@ -0,0 +1,95 @@
...
+type Service struct {
+	bootz.UnimplementedBootstrapServer
+	em EntityManager
+}
+
+func (s *Service) GetBootstrapRequest(ctx context.Context, req *bootz.GetBootstrapDataRequest) (*bootz.GetBootstrapDataResponse, error) {
+	if len(req.ChassisDescriptor.ControlCards) == 0 {
+		return nil, status.Errorf(codes.InvalidArgument, "request must include at least one control card")
+	}
+	lookup := &EntityLookup{
```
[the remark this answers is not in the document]
marcushines: i don't follow?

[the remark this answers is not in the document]
robshakir: It's not recommended to not use the field names here -- since if you add a new field then they aren't robust to it (it'll complain that it's not defined). There was a style guide update a while back.

client/README.md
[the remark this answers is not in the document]
robshakir: For the protocol version, yes -- but let's say you want to have a v2 of some of the client library which changes the API without changing the proto/service that is being referenced, doing this today would be difficult since you need to have the same tag for both the repo.

server/server.go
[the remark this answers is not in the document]
robshakir: Comments for packages are either ``// Package Foo` if this is a library or `// Binary <name of binary>` according to the style guide.

client/client.go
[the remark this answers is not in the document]
robshakir: SGTM.

server/service/service.go
robshakir: Suggest to add a comment that just explains this briefly.

client/README.md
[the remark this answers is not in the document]
marcushines: as we spoke we will align the go mod while also keeping the proto version

robshakir (approved):

PR #8, opened 2023-07-19
Add a simple fake client to Bootz

gmacf:
Fix gazelle rule and run `bazel run //:gazelle`


gmacf:
I'm new to gazele/bazel so I'm not sure if this is the correct thing to be doing. Let me know.

client/client.go
```diff
@@ -1,19 +1,163 @@
...
+			{
+				SerialNumber: "123A",
+				Slot:         1,
+				PartNumber:   "123A",
+			},
+			{
+				SerialNumber: "123B",
+				Slot:         2,
+				PartNumber:   "123B",
+			},
```
sam-homa, line 55: We should consider moving these to a single .conf file in future CLs. 

One potential benefit of doing this would be to only run server or client to test the other side (which is not mocked), that may require some more realistic data.

sam-homa: same goes for url and port below, those could be taken from either a single conf file or from command line or both

client/client.go
```diff
@@ -1,19 +1,163 @@
...
+		log.Exitf("Error calling GetBootstrapData: %v", err)
+	}
+
+	// Only check OC, OV and response signature if SecureOnly is set.
+	if secureOnly {
+		ov := resp.GetOwnershipVoucher()
+		if len(ov) == 0 {
+			log.Exitf("Received empty ownership voucher from server")
+		}
+		log.Infof("Verified Ownership Voucher")
```
sam-homa, line 115: this is good enough for the very first bare bone CL, however, we should consider verifying the signatures, OVs, etc. once we have the bare minimum at both client and server sides.

I am not sure if leaving TODO(s) is an accepted practice for these open-source repos. I'd guess it's probably acceptable for branches like this.

client/client.go
```diff
@@ -1,19 +1,163 @@
...
+	}
+
+	signedResp := resp.GetSignedResponse()
+	if secureOnly && signedResp.GetNonce() != nonce {
+		log.Exitf("GetBootstrapDataResponse nonce does not match")
+	}
+
+	for _, data := range signedResp.GetResponses() {
+		log.Infof("Received config for control card %v", data.GetSerialNum())
+		log.Infof("Downloading image %+v", data.GetIntendedImage())
```
sam-homa, line 136: I think two more steps here are to verify the hash of the image and to install the OS. We can leave TODO(s) to actually get those done in code if they are useful, even vendors can fill those as I am pretty sure they have the logic for them already (hopefully in Golang).

I think in general injecting small delays/sleeps (say a few seconds) between these steps can give the feel of the progress when you keep both client and server running in two terminals. Without sleeps in between, it'd just look like a quick dump at both sides.

sam-homa (approved):

client/client.go
gmacf: Done, I've added a validateArtifacts function to check OVs and OCs.

client/client.go
gmacf: Added a few TODOs and some sleeps to make this feel more realistic.

client/client.go
```diff
@@ -1,19 +1,307 @@
...
+	}
+
+	log.Infof("Validated ownership voucher signed by vendor")
+
+	// Verify the serial number for this OV
+	if parsedOV.OV.SerialNumber != serialNumber {
+		return fmt.Errorf("serial number from OV does not match request")
+	}
+
+	// Get the Pinned Domain Cert from the Ownership Voucher.
```
marcushines, line 101: maybe move this to a function just ease of reading

client/client.go
```diff
@@ -1,19 +1,307 @@
...
+	parsedOV := OwnershipVoucher{}
+	err = json.Unmarshal(p7.Content, &parsedOV)
+	if err != nil {
+		return err
+	}
+
+	// Create a CA pool for the device to validate that the vendor has signed this OV.
+	vendorCAPool := x509.NewCertPool()
+	if !vendorCAPool.AppendCertsFromPEM(rootCA) {
+		return fmt.Errorf("error adding vendor root CA to pool")
```
marcushines, line 85: you don't need "error" in the error string

client/client.go
```diff
@@ -1,19 +1,307 @@
...
+
+	// Verify the signature with the ownership certificate's public key. Currently only RSA keys are supported.
+	switch pub := ocCert.PublicKey.(type) {
+	case *rsa.PublicKey:
+		err = rsa.VerifyPKCS1v15(pub, crypto.SHA256, hashed[:], []byte(resp.GetResponseSignature()))
+		if err != nil {
+			return fmt.Errorf("signature not verified: %v", err)
+		}
+	default:
+		return fmt.Errorf("unsupported public key type")
```
marcushines, line 143: return fmt.Errorf("unsupported public key type: %T", pub)

client/client.go
```diff
@@ -1,19 +1,307 @@
...
+	}
+	return string(b), nil
 }
-func (c *Client) ReportStatus(ctx context.Context, req *bootz.ReportStatusRequest, opts ...grpc.CallOption) (*bootz.EmptyResponse, error) {
 
+func main() {
+	ctx := context.Background()
+	flag.Parse()
+
+	secureOnly := *bootMode == "SecureOnly"
```
marcushines, line 172: maybe add "" would be the other option here? or flip the flag to a bool?

client/client.go
gmacf: Is this just for the construction of the PDC or the rest of this function too?

client/client.go
gmacf: Done.

client/client.go
marcushines: just the pdc piece

client/ca.pem
```diff
@@ -0,0 +1,32 @@
+-----BEGIN CERTIFICATE-----
```
marcushines, line 1: maybe move to "cert directory" or something for "testdata"

client/client.go
```diff
@@ -1,19 +1,307 @@
...
+	"google.golang.org/protobuf/proto"
 )
 
-type Client struct {
-	bootz.BootstrapClient
+// Represents a 128 bit nonce.
+const nonceLength = 16
+
+var (
+	secureOnly = flag.Bool("secure_only", true, "Whether to start the emulated device in SecureOnly boot mode.")
```
marcushines, line 31: nit:

so for defaults generally follow the "go default" so in this case a bool is default false
so your flag would be 
--insecure_boot

client/client.go
```diff
@@ -1,19 +1,307 @@
...
+	CreatedOn                  string `json:"created-on"`
+	ExpiresOn                  string `json:"expires-on"`
+	SerialNumber               string `json:"serial-number"`
+	Assertion                  string `json:"assertion"`
+	PinnedDomainCert           string `json:"pinned-domain-cert"`
+	DomainCertRevocationChecks bool   `json:"domain-cert-revocation-checks"`
+}
+
+// pemEncodeCert adds the correct PEM headers and footers to a raw certificate block.
+func pemEncodeCert(contents string) string {
```
marcushines, line 51: nit:

return strings.Join([]string{"-----BEGIN CERTIFICATE-----", contents, "-----END CERTIFICATE-----"}, "\n")

client/client.go
```diff
@@ -1,19 +1,307 @@
...
+	// Create a new pool with this PDC.
+	pdcPool := x509.NewCertPool()
+	if !pdcPool.AppendCertsFromPEM([]byte(pdCPEM)) {
+		return err
+	}
+
+	// Parse the Ownership Certificate.
+	ocCert, err := certFromPemBlock(oc)
+	if err != nil {
+		log.Exitf("failed to parse certificate: %v", err)
```
marcushines, line 117: shouldn't exit from non main functions just return the error and let caller decide what to do

gmacf: Thanks, missed this one.

client/client.go
gmacf: Done.

client/client.go
gmacf: Done.

client/ca.pem
gmacf: Done.

client/client.go
gmacf: Done, I think. Let me know if more should be pulled out from this function.

marcushines (approved):

PR #9, opened 2023-08-03
Create a script to generate PDC, OC, OVs and Vendor CA certs.

gmacf:


testdata/generate.go
```diff
@@ -0,0 +1,235 @@
...
+func newCertificateAuthority(commonName string, org string) (string, string, error) {
+	// Create the certificate authority.
+	ca := &x509.Certificate{
+		SerialNumber: big.NewInt(int64(time.Now().Year())),
+		Subject: pkix.Name{
+			CommonName:   commonName,
+			Organization: []string{org},
+			Country:      []string{"US"},
+			Province:     []string{"CA"},
+			Locality:     []string{"Mountain View"},
```
sam-homa, line 67: These constants will work anyways, but I think we should move them to constants for now and later to move them all to a single constants package or .in file.

testdata/generate.go
```diff
@@ -0,0 +1,235 @@
...
+	pemBlock = strings.TrimSuffix(pemBlock, "\n-----END CERTIFICATE-----\n")
+	return pemBlock
+}
+
+// newOwnershipVoucher creates an OV for the device serial which is signed by the vendor's CA.
+func newOwnershipVoucher(serial string, pdcPem []byte, vendorCACert *x509.Certificate, vendorCAPriv *rsa.PrivateKey) (string, error) {
+	ov := OwnershipVoucher{
+		OV: OwnershipVoucherInner{
+			CreatedOn:        time.Now().String(),
+			ExpiresOn:        time.Now().Add(time.Hour * 24 * 365).String(),
```
sam-homa, line 117: ditto, would be good to separate out the time as a constant

testdata/README.md
```diff
@@ -2,6 +2,24 @@
...
+
+This is an x509 certificate/RSA keypair that represents the owner's Pinned Domain Cert.
+
+### oc_{priv|pub}.pem
+
+This is an x509 certificate/RSA keypair that represents the ownership certificate. It is signed by the PDC (or in some cases is the same as the PDC).
+
+### ov_x.txt
+
+Each of these files are base64-encoded, PKCS7 signed messages that represent the Ownership Voucher for that control card. It is signed by the Vendor CA and contains the PDC Cert.
```
sam-homa, line 25: it might be good to explain what control card means here. Something like what we have in our internal `xx testing` page:

```
 These actions happen at the level of Control Card which is the smallest unit to host and run the OS image. FFF (Fixed Form Factor) devices only have one unit of this per chassis, while MFF (Modular Form Factor) devices often have two (or more) of these per chassis. The term Control Card isn't used consistently across the industry. Each vendor have their own hardware architectures and hence their own terms for this bootstrapping unit including:

- Arista uses supervisors e.g. DCS-7280 is a single-supervisor (FFF) and DCS-7808 is a dual-supervisor (MFF).
- Cisco uses Routing Processors (RPs) e.g. CISCO-8202 is a single-RP (FFF) and CISCO-8808 is a dual-RP (MFF).
- Juniper uses Routing Engine (REs) e.g. PTX1000 (AKA PTX1k) is a single-RE (FFF) and PTX10008 is a dual-RE (MFF).
```

testdata/README.md
```diff
@@ -2,6 +2,24 @@
 
 This directory contains cryptographic artifacts necessary for the Bootz implementation. The files here are included as examples only and real implementations would need their own method of generating, storing and retrieving these.
 
-## ca.pem
+## Regenerating test data
 
-This is an x509 certificate that represents the device manufacturer's root (or some intermediate) CA. It is required that the ownership voucher passed from the Bootz server is signed by this CA.
\ No newline at end of file
+All certificates, private keys and OVs can be regenerated by running the `generate` binary. Ensure all text files are deleted from this directory before running to avoid conflicts.
```
sam-homa, line 7: It'd be useful to mention that these security artifacts should only be used for testing and MUST NOT be used in any production setup.

testdata/README.md
```diff
@@ -2,6 +2,24 @@
...
+
+The current files in this directory were generated using the command `./generate -vendor "Cisco" -owner "Google" -serials "123A,123B"`.
+
+### vendorca_{pub|priv}.pem
+
+This is an x509 certificate/RSA keypair that represents the device manufacturer's root (or some intermediate) CA. This is used to sign the OVs.
```
sam-homa, line 13: I think we should mention that self-signed certificates are used for convenience, but in production cases a reliable CA chain should be used instead.

sam-homa (approved):

testdata/generate.go
```diff
@@ -0,0 +1,245 @@
...
+
+// certFromPem converts a PEM-formatted byte slice to an x509 Certificate
+func certFromPem(contents []byte) (*x509.Certificate, error) {
+	pemBlock, _ := pem.Decode(contents)
+	return x509.ParseCertificate(pemBlock.Bytes)
+}
+
+// privateKeyFromPem converts a PEM-formatted byte slice to an RSA Private Key
+func privateKeyFromPem(contents []byte) (*rsa.PrivateKey, error) {
+	pemBlock, _ := pem.Decode(contents)
```
marcushines, line 61: this should return a nil, error on error case and then the next line will panic

you need to handle the error

testdata/generate.go
```diff
@@ -0,0 +1,245 @@
...
+	return x509.ParsePKCS1PrivateKey(pemBlock.Bytes)
+}
+
+// newCertificateAuthority creates a new CA for the chosen organization.
+// It returns a self-signed CA certificate as the first value, the associated private key as the second and any error as the third.
+func newCertificateAuthority(commonName string, org string) (string, string, error) {
```
marcushines, line 67: nit:
why return the strings vs. the data the you can let the caller deal with the object vs. it being string which can you can't really do anything else with

testdata/generate.go
```diff
@@ -0,0 +1,245 @@
...
+	defer f.Close()
+	b, err := f.Write(contents)
+	if err != nil {
+		return err
+	}
+	fmt.Printf("Wrote %d bytes to %v\n", b, filename)
+	return nil
+}
+
+func main() {
```
marcushines, line 173: nit:
for command line utilities we generally use cobra for all use cases as it allows everything to have the same look and feel 
for cli/ flags / testing

this is pretty trivial so agree it not critical

testdata/generate.go
```diff
@@ -0,0 +1,245 @@
...
+		log.Exit(err)
+	}
+
+	//Generate PDC.
+	fmt.Printf("Generating %v PDC cert and private key\n", *owner)
+	pdc, pdcPriv, err := newCertificateAuthority("Device Owner PDC", *owner)
+	if err != nil {
+		log.Exitf("unable to generate PDC: %v", err)
+	}
+	if err := writeFile([]byte(pdc), "pdc_pub.pem"); err != nil {
```
marcushines, line 205: so given you created these functions having to then "cohearse" things into other formats seems counter to design

marcushines (approved):

PR #12, opened 2023-08-15
Implement Entitymanager and populate security artifacts

gmacf:
This PR implements some basic entitymanager functionality and also reads the keypairs and OVs from disk.


server/entitymanager/entitymanager.go
```diff
@@ -2,38 +2,154 @@
...
+		BootMode: bootMode,
+	}
+	m.chassisInventory[l] = &e
+	return m
 }
 
 // New returns a new in-memory entity manager.
-func New() *InMemoryEntityManager {
-	return &InMemoryEntityManager{}
+func New(artifacts *service.SecurityArtifacts) *InMemoryEntityManager {
```
marcushines, line 149: we should have a clear workflow on how artifacts get loaded into the entity manager - i am seeing AddChasssis and AddControlCard

server/entitymanager/entitymanager.go
```diff
@@ -2,38 +2,162 @@
...
+		}
+		log.Infof("control card %v changed status from %v to %v", c.GetSerialNumber(), previousStatus, c.GetStatus())
+		m.controlCardStatuses[c.GetSerialNumber()] = c.GetStatus()
+	}
+	return nil
+}
+
+// Sign unmarshals the SignedResponse bytes then generates a signature from its Ownership Certificate private key.
+func (m *InMemoryEntityManager) Sign(resp *bootz.GetBootstrapDataResponse, serial string) error {
+	// Sign the response
```
sam-homa, line 93: nit: these comments are not providing much value imo.

server/entitymanager/entitymanager.go
```diff
@@ -2,38 +2,162 @@
...
+	}
+	if resp.GetSignedResponse() == nil {
+		return status.Errorf(codes.InvalidArgument, "empty signed response")
+	}
+	signedResponseBytes, err := proto.Marshal(resp.GetSignedResponse())
+	if err != nil {
+		return err
+	}
+	hashed := sha256.Sum256(signedResponseBytes)
+	sig, err := rsa.SignPKCS1v15(nil, priv, crypto.SHA256, hashed[:])
```
sam-homa, line 110: I think ECs are becoming more common and so I think it'd be a good idea to somehow detect the priv key type and support both algorithms. could be done as an improvement in a later PR though.

server/entitymanager/entitymanager.go
```diff
@@ -2,38 +2,162 @@
...
+	}
+	return nil
+}
+
+// Sign unmarshals the SignedResponse bytes then generates a signature from its Ownership Certificate private key.
+func (m *InMemoryEntityManager) Sign(resp *bootz.GetBootstrapDataResponse, serial string) error {
+	// Sign the response
+	block, _ := pem.Decode([]byte(m.artifacts.OC.Key))
+	if block == nil {
+		return status.Errorf(codes.Internal, "unable to decode OC private key")
```
sam-homa, line 96: The term "private key" is a little confusing to me. don't we only send public key in the OC (as a x509)?  Then the function argument is "priv", so I might be misunderstanding the flow here.

server/entitymanager/entitymanager_test.go
```diff
@@ -0,0 +1,291 @@
...
+					t.Skip()
+				}
+				t.Errorf("Sign() err = %v, want %v", err, test.wantErr)
+			}
+			signedResponseBytes, err := proto.Marshal(test.resp.GetSignedResponse())
+			if err != nil {
+				t.Fatal(err)
+			}
+			hashed := sha256.Sum256(signedResponseBytes)
+			sigDecoded, err := base64.StdEncoding.DecodeString(test.resp.GetResponseSignature())
```
sam-homa, line 159: nit: it feels to me that sign and verifySign should be in a package that could be used by both server and client code and unit tests

server/entitymanager/entitymanager.go
```diff
@@ -25,6 +26,7 @@ type InMemoryEntityManager struct {
 	// represents the current status of known control cards
 	controlCardStatuses map[string]bootz.ControlCardState_ControlCardStatus
 	artifacts           *service.SecurityArtifacts
+	mu                  sync.Mutex
```
marcushines, line 30: nit group the vars that mu protects below the mu

server/entitymanager/entitymanager_test.go
```diff
@@ -0,0 +1,291 @@
...
+	}
+}
+
+func TestResolveChassis(t *testing.T) {
+	tests := []struct {
+		desc    string
+		input   *service.EntityLookup
+		want    *service.ChassisEntity
+		wantErr bool
+	}{
```
marcushines, line 60: i meant for the format to be applied on all test cases

marcushines (approved):

PR #13, opened 2023-08-16
draft code for bootz server, dhcp plugin, and entity manager

mojiiba:
Hi @marcushines @gmacf 
Karim and I have been working on test bootzserver. I also noticed you are committing code that have some commonality. So, I am sharing this draft code to get early feedback from you and avoid duplicate work. 

@xw-g fyi


gmacf:
> Hi @marcushines @gmacf Karim and I have been working on test bootzserver. I also noticed you are committing code that have some commonality. So, I am sharing this draft code to get early feedback from you and avoid duplicate work.
> 
> @xw-g fyi

Hi @mojiiba, thanks for the contribution. It would be great to collaborate closely on this.  I'll take a look at this and give some comments. Could you also reach out to me (email is in profile) so we can discuss further?

client/client.go
```diff
@@ -220,7 +223,8 @@ func main() {
 	// 2. Bootstrapping Service
 	// Device initiates a TLS-secured gRPC connection with the Bootz server.
 	// TODO: Make this use TLS.
-	conn, err := grpc.Dial(bootzAddress, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	conn, err := grpc.Dial(bootzAddress, grpc.WithTransportCredentials(credentials.NewTLS(&tls.Config{InsecureSkipVerify: true})))
+	//conn, err := grpc.Dial(bootzAddress)
```
gmacf, line 227: Nit: Remove this comment?

dhcp/dhcp.go
gmacf, line 1: DHCP implementation looks good and we can plug this fairly easily

server/entitymanager/entitymanager.go
```diff
@@ -1,39 +1,278 @@
...
-	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/encoding/prototext"
 )
 
-// InMemoryEntityManager provides a simple in memory handler
-// for Entities.
-type InMemoryEntityManager struct{}
+type entityManager struct {
+	chassisConfigs []*epb.Chassis
```
gmacf, line 24: Perhaps could this be a map to make lookups easier? The key could be a struct like service.EntityLookup as it only contains manufacturer and serial.

client/client.go
athulma188, line 1: On line 144
		err = rsa.VerifyPKCS1v15(pub, crypto.SHA256, hashed[:], []byte(resp.GetResponseSignature()))

Should n't the conversion of signature from string to bytes be based on some encoding (base64 or hex)? I believe directly using the string representation of bytes generated by rsa.SignPKCS1v15 would give following error on the server
                rpc error: code = Internal desc = grpc: error while marshaling: string field contains invalid UTF-8

mojiiba:
will open separate pulls based on the issues.

Issue #20, opened 2023-08-24
Test APIs to control bootz server, dhcp server, and get device boot/dhcp status 


mojiiba:
We need to create APIs that allows starting/stopping DHCP/Bootz Server from tests. Also, we should add APIs to get device boot and dhcp ip lease status from tests.

mojiiba:
have submitted pull 28 for this.

PR #28, opened 2023-08-29
Add API and address config to enable testing bootz service in tests

mojiiba:
This pull addresses https://github.com/openconfig/bootz/issues/20. The unit tests will be completed before merge, but before doing  that just want to make sure you agree with the way i am doing it.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -32,7 +32,7 @@ const nonceLength = 16
 var (
 	verifyTLSCert = flag.Bool("verify_tls_cert", false, "Whether to verify the TLS certificate presented by the Bootz server. If false, all TLS connections are implicity trusted.")
 	insecureBoot  = flag.Bool("insecure_boot", false, "Whether to start the emulated device in non-secure mode. This informs Bootz server to not provide ownership certificates or vouchers.")
-	port          = flag.String("port", "", "The port to listen to on localhost for the bootz server.")
+	bootzAddress  = flag.String("address", "8008", "The [ip:]port to listen for the bootz server to connect. when ip is not given, the server will connect to localhost.")
 	rootCA        = flag.String("root_ca_cert_path", "../testdata/vendorca_pub.pem", "The relative path to a file containing a PEM encoded certificate for the manufacturer CA.")
 )
 
@@ -53,6 +53,16 @@ type OwnershipVoucherInner struct {
 // pemEncodeCert adds the correct PEM headers and footers to a raw certificate block.
 func pemEncodeCert(contents string) string {
 	return strings.Join([]string{"-----BEGIN CERTIFICATE-----", contents, "-----END CERTIFICATE-----"}, "\n")
+}
+
+// convert address to localhost when no ip is specefied
+func convertAddress(addr string) string {
+	items := strings.Split(addr, ":")
+	listenAddr := addr
+	if len(items) == 1 {
+		listenAddr = fmt.Sprintf("localhost:%v", addr)
+	}
+	return listenAddr
 }
 
 // validateArtifacts checks the signed artifacts in a GetBootstrapDataResponse. Specifically, it:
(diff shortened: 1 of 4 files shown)


gmacf:
I'm not a huge fan of putting test-only code in the Bootz serving code. Is there a way to instead change the public API to return the information we need to do tests?

mojiiba:
> I'm not a huge fan of putting test-only code in the Bootz serving code. Is there a way to instead change the public API to return the information we need to do tests?

Yes, I agree with you and the pr does that. The test in this pr is just a unit test for the APIs.

gmacf:
> > I'm not a huge fan of putting test-only code in the Bootz serving code. Is there a way to instead change the public API to return the information we need to do tests?
> 
> Yes, I agree with you and the pr does that. The test in this pr is just a unit test for the APIs.

Sorry, what I mean is I don't think we should be adding APIs to the Bootz server that will only be used for testing.

For example, you have added a method to the service that will only ever be used by tests"IsChassisConnected". Is there a way we can test if the chassis is connected by inspecting the response or error of the GetBootstrapData gRPC? Similar applies to getting boot status and resetting the server. IMO these are all things that should be evaluated by looking at the public API, not through a test-only one.

mojiiba:
no test has no way to access those info since the test is a separate entity and have no access to the message and logs transferred between the bootzserver and  the router, unless by adding APIs. We can also
(1) create a separate server along tests that only uses entity manager from the bootz repo and include all testing apis.  However, there will lead to duplicate code and impose a risk that test developers may introduce a behaviors that does not comply with the reference implementation. 
(2) may use grpc proxy between the router and bootzserevr to inspect the messages between the two, but this will make the testing complicated and probably flaky. Also, not sure how dhcp flow can be validated in this approach.

We can have an offline discussion on this if you want. 

@marcushines  what do u suggest?

gmacf:
> no, a test has no way to access those info since the test is a separate entity and have no access to the message and logs transferred between the bootzserver and the router, unless by adding APIs. We can also (1) create a separate server along tests that only uses entity manager from the bootz repo and include all testing apis. However, there will lead to duplicate code and impose a risk that test developers may introduce a behaviors that does not comply with the reference implementation. (2) may use grpc proxy between the router and bootzserevr to inspect the messages between the two, but this will make the testing complicated and probably flaky. Also, not sure how dhcp flow can be validated in this approach.
> 
> We can have an offline discussion on this if you want.
> 
> @marcushines what do u suggest?

Thanks for clarifying. I'll set up another meeting offline for us to discuss this.

server/service/service_test.go
```diff
@@ -0,0 +1,128 @@
+package service
```
melzhan, line 1: Please add a license header to all new files. I've just submitted a PR that adds them to all the other files so just look at one of them for reference.

mojiiba: sure, I will

client/client.go
```diff
@@ -55,6 +55,16 @@ func pemEncodeCert(contents string) string {
 	return strings.Join([]string{"-----BEGIN CERTIFICATE-----", contents, "-----END CERTIFICATE-----"}, "\n")
 }
 
+// convert address to localhost when no ip is specefied
```
melzhan, line 58: Comments that are complete sentences should be capitalized and punctuated like standard English sentences.

typo: specified

server/server.go
```diff
@@ -25,10 +25,20 @@ import (
 )
 
 var (
-	port              = flag.String("port", "", "The port to start the Bootz server on localhost")
+	bootzAddress      = flag.String("address", "8008", "The [ip:]port to listen for the bootz request. when ip is not given, the server will listen on localhost.")
 	artifactDirectory = flag.String("artifact_dir", "../testdata/", "The relative directory to look into for certificates, private keys and OVs.")
 )
 
+// convert address to localhost when no ip is specefied
```
melzhan, line 32: as above, typo and make full sentence

server/service/service_test.go
melzhan: Also note that packages must be preceded with a one line comment explaining the purpose of the package

mojiiba:
will open separate pulls for test apis

PR #31, opened 2023-09-04
Add license headers to all files

melzhan:
Following the guidelines:
- Add license headers to any files that can be added to (i.e. anything that takes the format of a source file and supports file comments). This would include things like YAML files.
- Include a CONTRIBUTING file which must include the Contributor License Agreement (CLA) section unaltered.


sam-homa (approved):
Are there any

PR #35, opened 2023-09-07
Entitymanager chassisInventory CRUD operations

LiamWalsh98:
Add methods to manage the chassis inventory of the entity manager (+ basic tests).

I have not opened any PRs to bootz before, so please let me know if there are any issues with the CLA

### DIFF
--- a/server/entitymanager/entitymanager.go
+++ b/server/entitymanager/entitymanager.go
@@ -354,3 +354,63 @@ func New(chassisConfigFile string) (*InMemoryEntityManager, error) {
 
 	return newManager, nil
 }
+
+func (m *InMemoryEntityManager) ReplaceDevice(chassis *service.EntityLookup, newChassis *epb.Chassis) error {
+	// chassis: old device, newChassis: new device
+
+	// todo: validate before replace
+	// todo: forward error from validateConfig
+
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	delete(m.chassisInventory, *chassis)
+
+	lookup := service.EntityLookup{
+		Manufacturer: newChassis.GetManufacturer(),
+		SerialNumber: newChassis.GetSerialNumber(),
+	}
+
+	m.chassisInventory[lookup] = newChassis
+
+	return nil
+}
+
+func (m *InMemoryEntityManager) DeleteDevice(chassis *service.EntityLookup) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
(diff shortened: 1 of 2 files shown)


melzhan:
Hi, looks like the Go/Static Analysis check is failing. Please fix before review :)

gmacf:
I think those static checks should start passing once you change your tests to use pointers to protobufs instead of the actual value.

LiamWalsh98:
Thank you for the comments! the most recent commit should address the failed check

server/entitymanager/entitymanager.go
```diff
@@ -354,3 +354,60 @@ func New(chassisConfigFile string) (*InMemoryEntityManager, error) {
 
 	return newManager, nil
 }
+
```
melzhan, line 357: Hi Liam, just letting you know we are working on importing this code into Google internally, which has strict code quality standards. I'll be leaving comments based on what needs to be updated such that the code can pass our internal linters.

You may notice some of the already submitted code already violates the comments I leave, I'll be going through and gradually fixing everything up.

server/entitymanager/entitymanager.go
```diff
@@ -354,3 +354,60 @@ func New(chassisConfigFile string) (*InMemoryEntityManager, error) {
...
+	delete(m.chassisInventory, *chassis)
+
+	lookup := service.EntityLookup{
+		Manufacturer: newChassis.GetManufacturer(),
+		SerialNumber: newChassis.GetSerialNumber(),
+	}
+
+	m.chassisInventory[lookup] = newChassis
+
+	// will return error when validation is added
```
melzhan, line 377: Comments that are complete sentences should be capitalized and punctuated like standard English sentences.

gmacf (approved):

server/entitymanager/entitymanager.go
LiamWalsh98: I mean... there is no direct object here so we can't call it a complete sentence.

But I got your point, hahaha. I'll fix up the comments. Thanks!

server/entitymanager/entitymanager.go
LiamWalsh98: Thank you, please let me know if there is anything else I need to change before this can go through!

server/entitymanager/entitymanager.go
melzhan: Cheers~ Also feel free to resolve conversations once you've addressed the comments :) I'll do it for you now and merge the PR. Thanks Liam~

PR #37, opened 2023-09-11
change the server to handle fixed chassis and modular when no serail …

mojiiba:
This pull change addresses https://github.com/openconfig/bootz/issues/38.


gmacf:
We're going to be reviewing https://github.com/openconfig/bootz/pull/39 instead of this one. Moji, could you change this to a draft or remove it?

PR #53, opened 2023-09-27
Remove dhcp_config field from inventory data as it does not exist in the proto message.

gmacf:
Also regenerate-files.sh script to build the entitymanager proto rules.


gmacf:
In a previous PR, dhcp_config was removed from the entity proto but the testdata textproto files were not updated to reflect that. This PR addresses that so that the server can build and run properly.

melzhan (approved):

PR #56, opened 2023-09-28
Refactor server to move logic out of main function.

gmacf:
This then allows us to create a startup test for the server.


melzhan (approved):
good luck

PR #66, opened 2023-10-04
Store serialized signed response instead of SignedResponse message.

gmacf:
To verify the signature of the signed response, we currently do this:
- Store the SignedResponse in the response proto
- Serialize the SignedResponse using proto.Marshal
-  Generate a signature from those bytes and store it in the response.

proto.Marshal is not a stable form of serialization which means we cannot guarantee that the signature is valid each time. Instead, the recommended approach is to store the serialized SignedResponse along with the signature instead of the actual SignedResponse message. This way, the client is responsible for verifying the signature and then unmarshaling.

### DIFF
--- a/server/entitymanager/entitymanager.go
+++ b/server/entitymanager/entitymanager.go
@@ -338,7 +338,7 @@ func (m *InMemoryEntityManager) Sign(resp *bpb.GetBootstrapDataResponse, chassis
 	if err != nil {
 		return err
 	}
-	sig, err := signature.Create(priv, signedResponseBytes)
+	sig, err := signature.Sign(priv, signedResponseBytes)
 	if err != nil {
 		return err
 	}
--- a/common/signature/signature.go
+++ b/common/signature/signature.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package signature provides methods to create and verify signatures.
+// Package signature provides methods to sign and verify signatures.
 package signature
 
 import (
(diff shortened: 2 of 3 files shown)


client/client.go
```diff
@@ -334,7 +328,11 @@ func main() {
 		}
 	}
 
-	signedResp := resp.GetSignedResponse()
+	var signedResp bpb.BootstrapDataSigned
+	if err := proto.Unmarshal(resp.GetSerializedSignedResponse(), &signedResp); err != nil {
```
yvetteyuanw, line 332: Not 100% sure if the unmarshal here would always work if the client and the server have a slightly different  Protobuf versions? Maybe include Charles as reviewer too? another option might be gRPC might have some libraries to handle this situation? other than this, LGTM.

yvetteyuanw (approved):

server/entitymanager/entitymanager.go
```diff
@@ -333,19 +333,12 @@ func (m *InMemoryEntityManager) Sign(resp *bpb.GetBootstrapDataResponse, chassis
...
-	log.Infof("Marshalling the response...")
-	signedResponseBytes, err := proto.Marshal(resp.GetSignedResponse())
-	if err != nil {
-		return err
-	}
-	log.Infof("Successfully serialized the response")
-
 	log.Infof("Calculating the sha256 sum to encrypt the response...")
-	hashed := sha256.Sum256(signedResponseBytes)
+	hashed := sha256.Sum256(resp.GetSerializedSignedResponse())
```
mojiiba, line 341: do we intend to make sha256 as the default?

mojiiba (approved):

client/client.go
melzhan: Protos should always be backwards compatible. (unless changed incorrectly). I'd advise we err on the side of keeping our protos backwards compatible. If the client and server protos aren't compatible we'd have pretty big problems...

chuckx: +1 to maintaining the protos in a backwards compatible fashion. [The protobuf best practices](https://protobuf.dev/programming-guides/dos-donts/) provide a set of guidelines for doing just that.

In fact, this PR is fixing a best practice violation, specifically [reliance on stable serialization, which is something you should never do](https://protobuf.dev/programming-guides/dos-donts/#serialization-stability).

The current implementation relies on unmarshaling (previously performed on line 126) to produce exactly the same bytes that were used when the signature was created. This is inherently unreliable since the signature is created by a different binary (i.e. the server) and at a different time (i.e. it may have had a different message definition).

Basically, the current implementation cannot reliably verify valid signatures, which this new approach fixes.

marcushines (approved):

PR #74, opened 2023-10-16
Replace direct proto field access with getters. 

gmacf:
This prevents nil pointer deferences in general.

### DIFF
--- a/server/service/service.go
+++ b/server/service/service.go
@@ -146,7 +146,7 @@ func (s *Service) GetBootstrapData(ctx context.Context, req *bpb.GetBootstrapDat
 	log.Infof("Response set")
 
 	// Sign the response if Nonce is provided.
-	nonce := reqGetNonce()
+	nonce := req.GetNonce()
 	if nonce != "" {
 		log.Infof("=============================================================================")
 		log.Infof("====================== Signing the response with nonce ======================")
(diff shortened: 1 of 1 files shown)


melzhan (approved):

server/service/service.go
```diff
@@ -145,11 +146,11 @@ func (s *Service) GetBootstrapData(ctx context.Context, req *bpb.GetBootstrapDat
 	log.Infof("Response set")
 
 	// Sign the response if Nonce is provided.
-	if req.Nonce != "" {
+	if req.GetNonce() != "" {
 		log.Infof("=============================================================================")
 		log.Infof("====================== Signing the response with nonce ======================")
 		log.Infof("=============================================================================")
-		resp.SignedResponse.Nonce = req.Nonce
+		resp.GetSignedResponse().Nonce = req.Nonce
```
marcushines, line 153: req.GetNonce() here as well

gmacf: In this case, GetNonce() returns the string by value, not by reference so it cannot be written to.

marcushines: resp.GetSignedResponse().Nonce = req.GetNonce()

marcushines (approved):

server/service/service.go
gmacf: Ah, sorry missed that.

PR #78, opened 2023-10-23
Add a new field 'serialized_bootstrap_data' and deprecate the old one.

gmacf:
The old field is still populated for backwards compatability but we now only guarantee the signature is valid for the serialized data.


client/client.go
```diff
@@ -121,13 +121,8 @@ func validateArtifacts(serialNumber string, resp *bpb.GetBootstrapDataResponse,
 	log.Infof("=============================================================================")
 	log.Infof("===================== Validating the response signature =====================")
 	log.Infof("=============================================================================")
-	log.Infof("Marshalling the response...")
-	signedResponseBytes, err := proto.Marshal(resp.GetSignedResponse())
-	if err != nil {
-		return err
-	}
 	log.Infof("Successfully serialized the response")
```
melzhan, line 124: Move this log under signature.Verify

proto/bootz.proto
```diff
@@ -141,7 +141,11 @@ message BootstrapDataSigned {
 }
 
 message GetBootstrapDataResponse {
```
melzhan, line 143: Just my idle musing: Since the proto is not yet being used for any RPC exchanges, isn't it ok to just remove the field? Why keep the proto backwards compatible when it doesn't need to be yet? Is there a benefit to giving vendors the option to use a more insecure format rather than only supporting the secure one?

gmacf: All vendors are using this field for testing and also for the upcoming image drop. If we deleted the field now it would break all integration and would delay progress quite a bit.

I'd rather get the initial testing working and slowly guide vendors to using the new field. Obviously we should make that a priority to do before we start ramping up fully.

client/client.go
gmacf: Done, thanks.

proto/bootz.proto
melzhan: okedoke~

It'd be good to send this to ckl@ to review if you haven't already~

melzhan (approved):

proto/bootz.proto
gmacf: Sounds good, we actually got feedback from him in https://github.com/openconfig/bootz/pull/66 but I accidentally closed that PR so making this one instead!

PR #82, opened 2023-10-25
Reconnect to server with the provided server trust cert.

gmacf:


client/client.go
```diff
@@ -342,6 +345,29 @@ func main() {
 		log.Infof("=============================================================================")
 	}
 
+	// Reconnect to the server with the provided server_trust_cert.
```
melzhan, line 348: Why do we do this? Did we not have the cert before? :O

gmacf: Yep, the first connection is untrusted. We can't make it trusted initially as the server's cert is self-signed. It's kinda documented in step 6 of https://github.com/openconfig/bootz#api-flow.

melzhan (approved):

PR #86, opened 2023-10-26
Use base64-encoded DER PinnedDomainCert in OwnershipVoucher instead of PEM format.

### LINKED ISSUE #69: Field should be byte instead of string
third_party/openconfig/bootz/common/ownership_voucher/ownership_voucher.go
[46] string

According to https://www.rfc-editor.org/rfc/rfc8366.html#section-5.1, this field is intended to be binary data.

Unless there's a reason I'm missing not to do so, this field should be []byte.

This would have the following benefits:

No need for RemovePemHeaders(). Instead of performing string manipulations of the encoded data, you would then use the encoding/pem package to more safely/reliably decode PEM and extract the byte sequence encoded within it.

No need to worry about manually encoding when serializing to JSON. When encoding/json marshals a []byte field, it automatically base64 encodes it. See https://pkg.go.dev/encoding/json#Marshal.

gmacf:
This removes the need to manually add and remove PEM headers.

closes #69


common/ownership_voucher/ownership_voucher.go
```diff
@@ -81,7 +73,7 @@ func New(serial string, pdcPem []byte, vendorCACert *x509.Certificate, vendorCAP
 			CreatedOn:        currentTime.String(),
 			ExpiresOn:        currentTime.Add(ovExpiry).String(),
 			SerialNumber:     serial,
-			PinnedDomainCert: RemovePemHeaders(string(pdcPem)),
+			PinnedDomainCert: pdcPem,
```
chuckx, line 76: My reading of https://www.rfc-editor.org/rfc/rfc8366.html#section-5.3 (specifically the definition of the `pinned-domain-cert` field on pages 12-13) is that this field should contain the DER encoded certificate, which is then base64 encoded when serialized to JSON. It makes no mention of PEM encoding the the DER encoded certificate.

If this field were actually populated with the DER bytes, then the following changes would need to be made (probably not exhaustive):

1. rename the function argument from `pdcPem` to `pdcDER`.
2. in `client.go` replace `pdcPool.AppendCertsFromPEM()` with `pdcPool.AppendCert()` after parsing the DER bytes with `x509.ParseCertificate()`.
3. in `generate.go`, populate the ownership vouchers with `pdc` instead of `pdcPem`
4. update tests accordingly

gmacf: Thanks, I've applied this to this PR.

gmacf: I've also changed PinnedDomainCert to be a string type again since it apparently should be base64 encoded.

melzhan (approved):

common/ownership_voucher/ownership_voucher.go
chuckx: DER is binary format for encoding ASN.1 data structures. It contains no implicit mechanisms for text encoding.

However, the examples ([section 5.2](https://www.rfc-editor.org/rfc/rfc8366.html#section-5.2)) depict the overall data structure encoded using JSON ([RFC8259](https://www.rfc-editor.org/rfc/rfc8259)). Also note that the data structure definition ([section 5.3](https://www.rfc-editor.org/rfc/rfc8366.html#section-5.3)) is using the YANG data modeling language ([RFC7950](https://www.rfc-editor.org/rfc/rfc7950), The `pinned-domain-cert` field is of type binary and that field type is defined to explicitly contain binary data, specifically a sequence of binary octets ([RFC7950, section 9.8](https://www.rfc-editor.org/rfc/rfc7950#section-9.8)).

In Go, the straightforward approach to representing such data is a byte slice.

Given a byte slice, the Go `encoding/json` package automatically encodes it using base64. From [the Marshal() documentation](https://pkg.go.dev/encoding/json#Marshal):

```
...[]byte encodes as a base64-encoded string...
```

[See this small Go Playground example demonstrating this.](https://go.dev/play/p/invuATCM6rA)

Basically, the struct field should be a byte slice. Then let the `encoding/JSON` package handle the base64 encoding, when (or if) it's necessary to serialize it to JSON.

Using a string field both makes the code more complex (requiring manual base64 encoding/decoding) and introduces storage overhead (the base64 encoded representation is ~35% larger than the original binary data).

gmacf: Thanks for clarifying. I've reverted this back to be a []byte and will rely on Json marshalling to do the base64 encoding as you mentioned.

PR #90, opened 2023-10-31
Return Owner Cert as a CMS structure instead of PEM encoding

gmacf:
From https://www.rfc-editor.org/rfc/rfc8572.html#section-3.2, this field should be a CMS structure, not a PEM encoding.

### DIFF
--- a/common/owner_certificate/owner_certificate_test.go
+++ b/common/owner_certificate/owner_certificate_test.go
@@ -16,56 +16,31 @@ package ownercertificate
 
 import (
 	"crypto/x509"
-	"encoding/pem"
 	"testing"
 
 	_ "embed"
-)
 
-var (
-	//go:embed testdata/oc_pub.pem
-	ocPub []byte
-	//go:embed testdata/pdc_pub.pem
-	pdcPub []byte
-	//go:embed testdata/oc_priv.pem
-	ocPriv []byte
+	artifacts "github.com/openconfig/bootz/testdata"
 )
 
 // Tests that the CMS structure can be created and that it can be verified with a PDC.
 func TestGenerateAndVerify(t *testing.T) {
-	block, _ := pem.Decode(ocPub)
-	if block == nil {
-		t.Fatalf("error decoding OC certificate")
-	}
-	ownerCert, err := x509.ParseCertificate(block.Bytes)
+	pdc, pdcPrivateKey, err := artifacts.NewCertificateAuthority("Pinned Domain Cert", "Google", "localhost")
 	if err != nil {
-		t.Fatal(err)
+		t.Fatalf("NewCertificateAuthority(): %v", err)
 	}
-	block, _ = pem.Decode(ocPriv)
-	if block == nil {
-		t.Fatalf("error decoding OC private key")
-	}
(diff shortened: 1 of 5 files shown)


gmacf:
@chuckx Please take a look if you have any spare time.  Thanks!

common/owner_certificate/owner_certificate_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+import (
+	"crypto/x509"
+	"encoding/pem"
+	"testing"
+
+	_ "embed"
+)
+
+var (
```
melzhan, line 25: I think this will break copybara when we import :< I wonder if there's a way to make it automatically pick up the right test files... 🤔

common/owner_certificate/owner_certificate_test.go
```diff
@@ -0,0 +1,71 @@
...
+	if err != nil {
+		t.Fatalf("error generating CMS: %v", err)
+	}
+	pdcPool := x509.NewCertPool()
+	pdcPool.AddCert(pdcCert)
+	_, err = Verify(cms, pdcPool)
+	if err != nil {
+		t.Fatalf("error verifying OC: %v", err)
+	}
+
```
melzhan, line 70: nit: remove extra line

common/owner_certificate/owner_certificate_test.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+	cms, err := GenerateCMS(ownerCert, ownerCertPrivateKey)
+	if err != nil {
+		t.Fatalf("error generating CMS: %v", err)
+	}
+	pdcPool := x509.NewCertPool()
+	pdcPool.AddCert(pdcCert)
+	_, err = Verify(cms, pdcPool)
+	if err != nil {
+		t.Fatalf("error verifying OC: %v", err)
```
melzhan, line 68: Couldn't find in the style guide what the correct format is. I've noticed some error strings have function name in them and some dont. Should we pick a style and stick with it?

common/owner_certificate/owner_certificate_test.go
gmacf: Which part specifically, the embedded files? We did a similar thing in https://github.com/openconfig/bootz/blob/6575eb87c559f913274c8e2182085a374b1f702f/common/ownership_voucher/ownership_voucher_test.go#L24.

melzhan (approved):

common/owner_certificate/owner_certificate_test.go
melzhan: no the list of testdata files. Buildozer isn't currently set up to automatically add them

gmacf: https://github.com/openconfig/bootz/pull/92 should help with this by removing all hardcoded data

common/owner_certificate/owner_certificate_test.go
gmacf: I think annotating the error with the function name is helpful here as it doesn't do that automatically. I've added it here and will do it for future error handling too.

Issue #105, opened 2023-12-04
De-duplicate protos


gmacf:
We currently have bootz.proto and entity.proto. The former is used for the gRPC API while the latter is mainly used for testing.

entity.proto is mostly a copy of bootz.proto and there's a lot of unfortunate duplication. We should deduplicate this and have bootz.proto as the main one.

mojiiba:
Yes, I agree. Please go ahead with the change and let me know if If you need help. 
Juts one request: We need to change the fp tests accordingly, so when your pr is ready, let us modify the tests and ensure all testing needs are addressed as part of the PR review.

PR #108, opened 2023-12-05
Include resolved bpb.Chassis in GetBootstrapData method.

gmacf:
This field is not currently used in the open source implementation but will be once we migrate to using a single proto message in https://github.com/openconfig/bootz/issues/105. 

Additionally, the open source implementation does not need this field yet but the internal one does.

### DIFF
--- a/server/entitymanager/entitymanager.go
+++ b/server/entitymanager/entitymanager.go
@@ -155,7 +155,7 @@ func populateBootConfig(conf *epb.BootConfig) (*bpb.BootConfig, error) {
 }
 
 // GetBootstrapData fetches and returns the bootstrap data response from the server.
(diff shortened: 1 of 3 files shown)


server/entitymanager/entitymanager.go
```diff
@@ -155,7 +155,7 @@ func populateBootConfig(conf *epb.BootConfig) (*bpb.BootConfig, error) {
 }
 
 // GetBootstrapData fetches and returns the bootstrap data response from the server.
-func (m *InMemoryEntityManager) GetBootstrapData(ctx context.Context, lookup *service.EntityLookup, controllerCard *bpb.ControlCard) (*bpb.BootstrapDataResponse, error) {
+func (m *InMemoryEntityManager) GetBootstrapData(ctx context.Context, ch *bpb.Chassis, lookup *service.EntityLookup, controllerCard *bpb.ControlCard) (*bpb.BootstrapDataResponse, error) {
```
melzhan, line 158: why add this if it's not used? Did you mean to replace line 163 below?

melzhan (approved):

PR #111, opened 2023-12-11
Add a new helper function "buildEntityLookup"

gmacf:
Currently there's no consistent way of looking up modular or fixed form factor chassis'. This is the first step in having a single entity lookup builder.

Note: This function isn't used yet, this will be plugged into service.go in a future PR.

### DIFF
--- a/server/service/service.go
+++ b/server/service/service.go
@@ -58,13 +58,54 @@ type SecurityArtifacts struct {
 	TLSKeypair *tls.Certificate
 }
 
-// EntityLookup provides a way to resolve chassis and control cards
-// in the EntityManager.
+// EntityLookup is used to resolve the fields of an active control card to a chassis.
+// For fixed form factor devices, the active control card is the chassis itself.
 type EntityLookup struct {
+	// The manufacturer of this control card or chassis.
 	Manufacturer string
+	// The serial number of this control card or chassis.
 	SerialNumber string
-	PartNumber   string
-	IPAddress    string
+	// The hardware model/part number of this control card or chassis.
+	PartNumber string
+	// The reported IP address of the management interface for this control
+	// card or chassis.
+	IPAddress string
+	// Whether this chassis appears to be a modular device.
+	Modular bool
+}
+
+// buildEntityLookup constructs an EntityLookup object from a bootstrap request.
+func buildEntityLookup(ctx context.Context, req *bpb.GetBootstrapDataRequest) (*EntityLookup, error) {
+	peerAddr, err := peerAddressFromContext(ctx)
+	if err != nil {
+		return nil, err
+	}
+	activeControlCardSerial := req.GetControlCardState().GetSerialNumber()
+	var partNumber string
+	var modular bool
+	if len(req.GetChassisDescriptor().GetControlCards()) == 0 {
+		modular = false
+		partNumber = req.GetChassisDescriptor().GetPartNumber()
+	} else {
+		modular = true
+		for _, card := range req.GetChassisDescriptor().GetControlCards() {
+			if card.GetSerialNumber() == activeControlCardSerial {
+				partNumber = card.GetPartNumber()
+				break
+			}
+		}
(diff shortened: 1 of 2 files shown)


server/service/service.go
```diff
@@ -58,13 +58,54 @@ type SecurityArtifacts struct {
...
+	peerAddr, err := peerAddressFromContext(ctx)
+	if err != nil {
+		return nil, err
+	}
+	activeControlCardSerial := req.GetControlCardState().GetSerialNumber()
+	var partNumber string
+	var modular bool
+	if len(req.GetChassisDescriptor().GetControlCards()) == 0 {
+		modular = false
+		partNumber = req.GetChassisDescriptor().GetPartNumber()
```
melzhan, line 88: would this just be an empty string?

server/service/service.go
```diff
@@ -58,13 +58,54 @@ type SecurityArtifacts struct {
...
+		modular = true
+		for _, card := range req.GetChassisDescriptor().GetControlCards() {
+			if card.GetSerialNumber() == activeControlCardSerial {
+				partNumber = card.GetPartNumber()
+				break
+			}
+		}
+	}
+	if partNumber == "" {
+		return nil, status.Errorf(codes.InvalidArgument, "active control card with serial %v not found in chassis descriptor", activeControlCardSerial)
```
melzhan, line 99: NotFound?

server/service/service.go
```diff
@@ -58,13 +58,54 @@ type SecurityArtifacts struct {
...
+	Modular bool
+}
+
+// buildEntityLookup constructs an EntityLookup object from a bootstrap request.
+func buildEntityLookup(ctx context.Context, req *bpb.GetBootstrapDataRequest) (*EntityLookup, error) {
+	peerAddr, err := peerAddressFromContext(ctx)
+	if err != nil {
+		return nil, err
+	}
+	activeControlCardSerial := req.GetControlCardState().GetSerialNumber()
```
melzhan, line 83: Should we null check this value?

server/service/service.go
gmacf: Do you mean line 88? (GitHub does a bad job at showing what the comment is anchored to). 

It shouldn't be empty, the proto says that you can either populate the control cards (for modular) or populate chassis descriptor fields (for fixed form factor).

server/service/service.go
gmacf: NotFound implies that the server could not find something (e.g. server's fault). I think invalid argument is more correct here as it is actually the user/client that hasn't filled in the request properly.

server/service/service.go
melzhan: I would advise either adding a check that it's not empty if it shouldn't be, or at least adding a comment that it shouldn't be.

server/service/service.go
melzhan: oke sgtm

server/service/service.go
gmacf: Yep, fixed that now.

server/service/service.go
```diff
@@ -58,13 +58,57 @@ type SecurityArtifacts struct {
...
+	if activeControlCardSerial == "" {
+		return nil, status.Errorf(codes.InvalidArgument, "no active control card serial provided")
```
melzhan, line 85: For completeness, consider adding a corresponding test

server/service/service.go
gmacf: We check if part number is set for modular or fixed form factor on line 101. So at least one of these places must have it populated.

melzhan: okedoke

melzhan (approved):

server/service/service.go
gmacf: I did in https://github.com/openconfig/bootz/pull/111/commits/e7b93b17da56d2375ff7c1d2d2a884f17413540f :)

PR #125, opened 2024-01-11
Update bootz.proto

marcushines:
Change package name to be correct


xw-g (approved):

gmacf:
What's the reason for this change?

This is going to break Bootz as it will change the gRPC service name exported by the server. The software versions used by our test devices will contain the old gRPC service name and won't be able to communicate.

PR #128, opened 2024-02-27
Add support for rfc8572 DHCP option encoding.

### LINKED ISSUE #116: DHCP Option encoding
Hi,
I have concern with the way Bootz is using DHCP options 143 for v4 and 136 for v6.
Since this is IETF standards based option, anyone who wants to use this option should
be in conformance with RFC. 

This option is standardized by RFC 8572, section 8. In future, when DHCP servers starts 
adding support for this option, they will follow IETF since it is standardized way.
So, bootz should also align with this uniform encoding so that both Secure-ZTP and
Bootz can work with same DHCP option type.

how it should be:
(full) option length should follow the type-143. Size 1-bytes.
(Section-8.1 of RFC 8572)

             0                             1
             0  1  2  3  4  5  6  7  8  9  0  1  2  3  4  5
            +--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+
            |   option-code (143)   |     option-length     |
            +--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+
            .                                               .
            .    bootstrap-server-list (variable length)    .
            .                                               .
            +--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+

            * option-code: OPTION_V4_SZTP_REDIRECT (143)
            * option-length: The option length in octets.
            * bootstrap-server-list: A list of servers for the
               client to attempt contacting, in order to obtain
               further bootstrapping data, in the format shown
               in [Section 8.3](https://datatracker.ietf.org/doc/html/rfc8572#section-8.3).

Then server-list is encoded like this (Section 8.3 of RFC 8572):

Both of the DHCPv4 and DHCPv6 options defined in this section encode
   a list of bootstrap server URIs.  The "URI" structure is a DHCP
   option that can contain multiple URIs (see [RFC7227], Section 5.7).
   Each URI entry in the bootstrap-server-list is structured as follows:

    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-...-+-+-+-+-+-+-+
    |       uri-length              |          URI                  |
    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-...-+-+-+-+-+-+-+

    * uri-length: 2 octets long; specifies the length of the URI data.
    * URI: URI of the SZTP bootstrap server.

Since length is 2 bytes, it is will uint-short. This should be encoded in htons() format.
Also, this enables sending multiple servers as list for load balancing and redundancy.

Yes, it looks a bit silly since full option length is 8 bits thus max 255 bytes while each 
URL length field is 16-bit which is sub-part of 255 bytes. But "this is the way" :-)

Not sure if following code from plugins/bootz/bootz.go is doing same as above. But wanted to 
make sure we are doing right thing.

       ztpV4Opt = &dhcpv4.Option{
                Code:  dhcpv4.GenericOptionCode(OPTION_V4_SZTP_REDIRECT),
                Value: dhcpv4.String(url.String()),
        }

Thanks,
Katta

gmacf:
Closes #116

### DIFF
--- a/dhcp/dhcp.go
+++ b/dhcp/dhcp.go
@@ -40,8 +40,8 @@ const confTemplate = `
 server6:
    plugins:
      - server_id: LL {{ .IntfMacAddr }}
-     {{ if .BootzURL }}
-     - bootz: {{ .BootzURL }}
+     {{ if .BootzURLs }}
+     - bootz: {{ .BootzURLs }}
      {{ end }}
      {{ if .DNSv6 }}
      - DNS: {{ .DNSv6 }}
@@ -53,8 +53,8 @@ server4:
   plugins:
     - lease_time: 3600s
     - server_id: {{ .IntfIPAddr }}
-    {{ if .BootzURL }}
-    - bootz: {{ .BootzURL }}
+    {{ if .BootzURLs }}
+    - bootz: {{ .BootzURLs }}
     {{ end }}
     {{ if .DNSv4 }}
     - DNS: {{ .DNSv4 }}
@@ -77,7 +77,7 @@ type Config struct {
 	Interface  string
 	DNS        []string
 	AddressMap map[string]*Entry
-	BootzURL   string
+	BootzURLs  []string
 }
 
 // Entry represents a dhcp record.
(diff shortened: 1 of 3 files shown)


gmacf:
@sam-homa the logic here for encoding the DHCP options  is basically taken from our internal DHCP server. I figured you're the most familiar with this.

sam-homa (approved):

PR #133, opened 2024-03-15
Remove all bazel build files

melzhan:


marcushines (commented):
why are removing bazel build files

melzhan:
Bazel isn't being used to build any of the repo and as a result no one's been maintaining it. Currently the bazel code doesn't build, and we had some vendors get confused. Figured we'd remove it to mitigate confusion.

gmacf (approved):

PR #142, opened 2024-07-22
Add SkipEnrollment flag to control the boot config.

sam-homa:
Give enrollz signal to bootz to add/remove additional knobs from/to the boot config.

### DIFF
--- a/server/service/service.go
+++ b/server/service/service.go
@@ -100,6 +100,9 @@ type Chassis struct {
 	BootloaderPasswordHash string
 	// The name of the fabric this chassis belongs to.
 	FabricName string
+	// Indicates whether enrollment will be skipped later in the install process.
+	// Some parts of boot config may be different depending on the value of this flag.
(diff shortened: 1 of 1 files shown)


server/service/service.go
```diff
@@ -100,6 +100,9 @@ type Chassis struct {
 	BootloaderPasswordHash string
 	// The name of the fabric this chassis belongs to.
 	FabricName string
+	// Indicates whether enrollment will be skipped later in the install process.
+	// Some parts of boot config may be different depending on the value of this flag.
```
melzhan, line 104: Can you add a todo to either remove this, or refactor this so it makes more sense?

melzhan (approved):

server/service/service.go
sam-homa: Added a TODO and a bug to track it.

gmacf (approved):

PR #147, opened 2025-01-14
Update build to support Bazel 8/bzlmod

bstoll:
This PR updates this repo to support bzlmod/Bazel8 and attempts to generate BUILD files with Gazelle.

This PR is flagged as a draft because it pins openconfig_gnmi and openconfig_gnsi to my personal git branch. https://github.com/openconfig/gnmi/pull/192 and https://github.com/openconfig/gnsi/pull/220 need to be merged and uploaded to the Bazel Central Registry before this PR can be merged.


dplore (approved):
LGTM.  I dunno if @marcushines wants to set me as a maintainer instead of or in addition to him.  I'll let him comment.

bstoll:
The local pins to openconfig_gnsi and openconfig_gnmi are removed and now depend on the main Bazel Central Registry releases.

Gazelle targets are included but right now Gazelle will generate some bad/broken BUILD rules. Fixing this may require moving some files and changing target names - this would be a bigger breaking change.  This PR is already big enough just getting bzlmod support in so I would prefer to tackle that issue later.

PR #150, opened 2025-03-03
Add streaming rpc for bootz for support of standard tls certs rather

marcushines:
than forcing use of iDevID


gmacf:
Internally, we configure the IDevID requirement  in the TLS handshake at the server level. I don't know of a way (yet) to relax this requirement to only apply to some RPC methods.

So even if we were to implement this in the current state, the server still has a single TLS configuration that would enforce the use of IDevID for BootstrapStream() and all other RPC methods.

I would need to take a look at the TLS/Go libraries available to us internally first to see if it's even feasible to do this.

gmacf:
> it would be applied to the tls profile which be applied to add services not just some rpc's

I'm not talking about the device. I'm talking about how we implement this streaming RPC on the Bootz server side. 

> This bypasses the need to require the iDevID cert to be sent as part of the initial gRPC TLS handshake. 

This IDevID cert in TLS handshake requirement applies to the whole Bootz gRPC server. It's not possible from what I can see to selectively require an IDevID in the TLS connection for only some methods.

A workaround to this is that we would spin up a streaming-only Bootz server that only implements this method and doesn't require IDevID.

proto/bootz.proto
```diff
@@ -43,6 +43,21 @@ service Bootstrap {
   // out-of-service state. The device should validate the server's identity
   // against the server_trust_cert it obtained in GetBootstrappingDataResponse.
   rpc ReportStatus(ReportStatusRequest) returns (EmptyResponse) {}
+
+  // BootstrapStream provides the streaming implementation of the Bootz
+  // service. This flow allows for both server and client validation via
+  // nonces and certficate validation. This allows the service to start with
```
robshakir, line 49:
```suggestion
  // nonces and certificate validation. This allows the service to start with
```

robshakir (approved):
Thanks for this -- makes sense. Thanks for explaining this to me offline :-)

proto/bootz.proto
```diff
@@ -54,11 +69,43 @@ message GetBootstrapDataRequest {
...
+
+// BootstrapStreamRequest is used to both build send chassis information
+// as well as build a trust relationship if the iDevID cert is not used
+// during TLS setup.
+message BootstrapStreamRequest {
+  oneof type {
+    GetBootstrapDataRequest bootstrap_request = 1;
+    // Server provided nonce signed with the IAK from the device.
+    // For non-iDevID systems the Nonce will be signed with:
+    //   TPM 1.2: EK
```
nmahabaleshwar, line 91: EK/PPK cannot be used for signing. We need a challenge/response scheme which is different for IDevId vs non-IDevId systems. Below is our proposal,

1. The device sends either an IDevId certificate or an EK/PPK public key (if IDevId certificate is not present) to the Bootz server.
2. Now the server has 2 different challenge mechanisms based on what was received from the device,
    2a. IDevId certificate: Send a plaintext nonce which the device is expected to sign and return using the IDevID key.
    2b: EK/PPK: Send an encrypted nonce using the EK/PPK public key.
3. Similarly the device has 2 different response mechanisms based on what the Bootz server sent,
     3a. Plaintext nonce: Send sign(nonce, IDevId) back to the Bootz server
     3b. Encrypted nonce: Decrypt using the EK/PPK and send back the nonce back to the Bootz server.
4. If the Bootz server is satisfied with the response from the device, it sends the BootstrapDataResponse.

```
message GetBootstrapDataRequest {
   message Identity {
      string idevid_cert = 1;
      string ek_pub = 2;
      string ppk_pub = 3;
   }
  ChassisDescriptor chassis_descriptor = 1;
  ControlCardState control_card_state = 2;
  string nonce = 1001;
  Identity identity = 1003;
}

message BootstrapStreamRequest {
  message Response {
     oneof type {
       bytes nonce_signed = 1;
       string nonce = 2;
     }
  }
  oneof type {
    GetBootstrapDataRequest bootstrap_request = 1;
    Response response = 2;
  }
}

message BootstrapStreamResponse {
  message Challenge {
    oneof type {
      string nonce = 1;
      bytes nonce_encrypted = 2;
    }
  }
  oneof type {
    Challenge challege = 1;
    GetBootstrapDataResponse bootstrap_response = 2;
  }
}
```

Note that I am proposing IDevId certificate as a PEM encoded string to be uniform across Bootz/Attestz.

marcushines: i thought we wanted to get the pub's from ovgs?
i guess sending the pub's over the channel is "kinda okay"

proto/bootz.proto
```diff
@@ -57,6 +86,54 @@ message GetBootstrapDataRequest {
...
+// during TLS setup.
+message BootstrapStreamRequest {
+  message Response {
+    oneof type {
+      // Nonce to be signed with IAK from the device for iDevID based systems.
+      bytes nonce_signed = 1;
+      // For non-iDevID systems the encrypted_nonce will be decrypted via:
+      //   TPM 1.2: EK
+      //   TPM 2.0: PPK
+      string nonce = 2;
```
nmahabaleshwar, line 107: Should we leave this as bytes ?

marcushines: since the getbootstraprequest contains a nonce  as a string i think we should keep the types consistent

nmahabaleshwar (approved):

PR #160, opened 2025-06-10
Add ReportStatus messages as an allowed type in the BootstrapStream flow

gmacf:
This PR makes the ReportStatus message compatible with the new challenge-response authentication method for the BootstrapStream RPC.

This enables the device to report to the Bootz server when it has finished bootstrapping. Without this functionality, Bootz server has no reliable way of knowing when to initiate the next step of the turnup process which is TPM enrollment.

### DIFF
--- a/proto/bootz.proto
+++ b/proto/bootz.proto
@@ -66,20 +66,18 @@ service Bootstrap {
 // If identity is set one of the types must be set or an error is
 // returned.
 message Identity {
-  message Identity {
-    oneof type {
-      // idevid_cert will be validated via OVGS and the vendor CA bundle.
-      // Additional it will be used to validate the nonce_signed.
-      // The format is a base64 encoding of an x509 DER certificate.
-      string idevid_cert = 1;
-      // One of the key types must be set for non-iDevID based systems.
-      // The public keys will be used to encrypt the nonce and validate
-      // that the keys match the data in OVGS. The decrypted
-      // nonce will be returned to validate the device identity.
-      bool ek_pub = 2;
-      bool ppk_pub = 3;
-    }
-  } 
+  oneof type {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -460,6 +460,20 @@ the ownership voucher and ownership certificate.
         1. A reboot may be performed, if required.
     5. The device will then apply the configuration
         1. A reboot may be performed, if required.
+6. BootstrapStreamRequest.report_status_request
+    1. The device sends a ReportStatusRequest message to the bootz-server.
+    2. If bootstrapping is successful, the device should report all control cards as `CONTROL_CARD_STATUS_INITIALIZED` and `BOOTSTRAP_STATUS_SUCCESS`.
+    3. If bootstrapping is not successful, the device *may* report status as `BOOTSTRAP_STATUS_FAILURE` and then should restart the process from step 1.
+    4. If the device terminated the previous stream (due to a reboot or to apply a configuration)
```
melzhan, line 467: nit: maybe 4a. and 4b. since they're connected to the same point?

proto/bootz.proto
```diff
@@ -60,9 +60,12 @@ service Bootstrap {
...
 
-// Initial request from device to the bootz-server. Contains a detailed
-// description of the chassis and its constituent control cards.
-message GetBootstrapDataRequest {
+// Describes how the device will prove its identity to the Bootstrap server
+// during the BootstreapStream() RPC.
+// For systems which support iDevID that will be the preferred method.
+// If identity is set one of the types must be set or an error is
+// returned.
+message Identity {
```
melzhan, line 68: why nest an identity message inside an identity message?

melzhan (approved):

PR #201, opened 2025-08-27
implement TPM2.0 with idevid streaming bootz service BootstrapRequest and BootstrapResponse

seal6363:


server/service/service.go
```diff
@@ -269,29 +275,204 @@ func (s *Service) ReportStatus(ctx context.Context, req *bpb.ReportStatusRequest
...
 func (s *Service) BootstrapStream(stream bpb.Bootstrap_BootstrapStreamServer) error {
+	ctx := stream.Context()
+	clientHasCert := hasClientCert(ctx)
+
+	var deviceID string
+	var chassis *epb.Chassis
+	var activeControlCard string
+	currentState := stateInitial
+
+	if clientHasCert {
```
gmacf, line 323: BootstrapStream doesn't use any client certs so you can remove this logic.

server/service/service.go
```diff
@@ -269,29 +275,204 @@ func (s *Service) ReportStatus(ctx context.Context, req *bpb.ReportStatusRequest
...
 		case *bpb.BootstrapStreamRequest_BootstrapRequest:
-			// TODO: process request
+			if currentState != stateInitial {
+				return status.Errorf(codes.FailedPrecondition, "received BootstrapRequest in unexpected state: %v", currentState)
+			}
+			bootstrapReq := req.BootstrapRequest
+			identity := bootstrapReq.GetIdentity()
+
+			if clientHasCert {
+				// --- TPM 2.0 with IDevID Flow ---
```
gmacf, line 361: Similar to above, the IDevID flow in BootstrapStream doesn't rely on any client cert. You need to check Identity() in the Bootstrap request instead

PR #204, opened 2025-09-01
implement BootstrapStream request for tpm 2.0 with idevid

seal6363:


server/service/service.go
```diff
@@ -269,29 +288,165 @@ func (s *Service) ReportStatus(ctx context.Context, req *bpb.ReportStatusRequest
...
+	stateTPM12EKIdentitySent
+	// TPM 2.0 states
+	stateTPM20CSRRequested
+	stateTPM20NonceSent
+	// Common state
+	stateAttested
+)
+
+// Function to check if a client cert was presented
+func hasClientCert(ctx context.Context) bool {
```
gmacf, line 304: I think it's fine to remove this check. Reason is the Bootz server performs this validation already (e.g. the interceptor logic you wrote).

server/service/service.go
```diff
@@ -269,29 +288,165 @@ func (s *Service) ReportStatus(ctx context.Context, req *bpb.ReportStatusRequest
...
+	if hasClientCert(ctx) {
+		return status.Errorf(codes.FailedPrecondition, "BootstrapStream is not for clients presenting client certificates in TLS; use GetBootstrapData/ReportStatus RPCs instead")
+	}
+
+	var deviceID string
+	session := &streamSession{currentState: stateInitial}
+	var err error
+
+	// --- Initial Message ---
+	in, err := stream.Recv()
```
gmacf, line 328: Would it be possible to remove this single call to stream.Recv() and instead just have everything in a constant loop (e.g. on line 427().

server/service/service.go
```diff
@@ -157,6 +162,20 @@ type EntityManager interface {
 type Service struct {
 	bpb.UnimplementedBootstrapServer
 	em EntityManager
+	// Session cache for nonces or other state for non-IDevID flows
+	sessions sync.Map // map[string]*streamSession
```
gmacf, line 166: I don't think you need to keep the session state in the Service level. It only needs to exist for the lifetime of the stream. In other words, each BootstrapStream() call will handle exactly one gRPC stream and should be deleted when the stream ends.

seal6363: good to know. makes this simpler.

server/service/service.go
```diff
@@ -269,29 +284,121 @@ func (s *Service) ReportStatus(ctx context.Context, req *bpb.ReportStatusRequest
...
+				}
+				if err := stream.Send(challengeResp); err != nil {
+					return err
+				}
+				session.currentState = stateTPM20NonceSent
+				log.Infof("Sent nonce challenge to IDevID device: %s", deviceID)
+
+			case *bpb.Identity_EkPub:
+				return status.Errorf(codes.Unimplemented, "TPM 2.0 without IDevID flow not fully implemented")
+			default:
```
gmacf, line 392: Could you also include the other flows here as unimplemented? E.g. ek_ppk_pub and ppk_pub.

gmacf: That was a recent addition to the proto so you might need to pull the latest changes for that.

gmacf (approved):

seal6363:
implement BootstrapStream request for tpm 2.0 with idevid

PR #208, opened 2025-09-08
implement boostrap stream response

seal6363:


server/service/service_test.go
```diff
@@ -244,16 +247,16 @@ func createTestClient(t *testing.T, addr string, tlsClientCreds credentials.Tran
...
-	// For simplicity, using insecure here, as the server logic *rejects* client certs.
-	return insecure.NewCredentials()
-}
-
 func TestBootstrapStream(t *testing.T) {
-	goodCert := base64.StdEncoding.EncodeToString(createTestCertificate(t, "test-device", "test-serial-123"))
+	devicePrivKey, goodCertDER := createTestCertificate(t, "test-device", "test-serial-123")
+	goodCert := base64.StdEncoding.EncodeToString(goodCertDER)
+	// Create a bootloader config struct for the test case.
+	bootloaderConfig, err := structpb.NewStruct(map[string]interface{}{
```
gmacf, line 254: The bootloader config is actually not a commonly implemented field for vendors. For testing here, I'd recommend instead using "vendor_config". This is saves having to make that Struct type.

seal6363: good catch.

server/service/service.go
```diff
@@ -404,7 +405,64 @@ func (s *Service) BootstrapStream(stream bpb.Bootstrap_BootstrapStreamServer) er
...
+			publicKey, ok := session.idevidCert.PublicKey.(*rsa.PublicKey)
+			if !ok {
+				return status.Errorf(codes.InvalidArgument, "public key from idevid_cert is not of type RSA")
+			}
+			hasher := crypto.SHA256.New()
+			hasher.Write(session.nonce)
+			hashedNonce := hasher.Sum(nil)
+
+			if err := rsa.VerifyPKCS1v15(publicKey, crypto.SHA256, hashedNonce, signedNonce); err != nil {
+				return status.Errorf(codes.PermissionDenied, "nonce signature verification failed: %v", err)
```
gmacf, line 431: Permission denied probably isn't the right error code here. Internal or invalid argument is better.

server/service/service.go
```diff
@@ -404,7 +405,64 @@ func (s *Service) BootstrapStream(stream bpb.Bootstrap_BootstrapStreamServer) er
 			}
 
 		case *bpb.BootstrapStreamRequest_Response_:
-			// TODO: process response
+			log.Infof("Received Response from %s", deviceID)
+			if session.currentState != stateTPM20NonceSent {
+				return status.Errorf(codes.FailedPrecondition, "unexpected state %v for device %s, expecting nonce response", session.currentState, deviceID)
```
gmacf, line 410: For this, I would use invalid argument. Failed precondition implies the server is in the wrong state but in this context it's because the device is in the wrong state.

server/service/service.go
seal6363: done

PR #217, opened 2025-09-15
add bootstrap stream report status

seal6363:


server/service/service.go
```diff
@@ -384,7 +384,48 @@ func (s *Service) BootstrapStream(stream bpb.Bootstrap_BootstrapStreamServer) er
...
+			// knowing which device is reporting status, making it impossible to issue a
+			// secure re-challenge.
+			//
+			// By returning `FailedPrecondition`, we reject the ambiguous request and
+			// force the client to restart the entire authentication flow. This ensures
+			// that every new connection is fully and securely authenticated from the
+			// beginning, starting with a `BootstrapRequest` that contains the
+			// necessary device identity.
+
+			if session.chassis == nil {
```
gmacf, line 408: While it is true that we can't accept a ReportStatus request if we have not established trust, we shouldn't immediately return an error. Instead we should perform the same challenge response process as before. Consider this case:

Device performs BootstrapStream(), authenticates successfully and applies bootstrap data.  Since it's already authenticated and has the stream open, it sends a final ReportStatus message saying that it's going to reboot to apply the config.

At this point, it will close the stream (and therefore our server will lose track of the session). When it boots up again, it will send a ReportStatus message saying it's done in a new stream and again perform the same challenge-response process as before. See README.me for where this is documented:

"If the device terminated the previous stream (due to a reboot or to apply a configuration)
The server responds with a BootstrapStreamResponse.challenge to re-authenticate with the device (see step 7).
If the device did not terminate the previous stream
the server responds with a ReportStatusResponse (See step 9)"

gmacf: Also, the comments above this don't really make sense.  

> By returning `FailedPrecondition`, we reject the ambiguous request and force the client to restart the entire authentication flow.

This doesn't really work because we're not giving the client a chance to restart the auth flow. The current code is just always returning failed precondition where it should be responding with a challenge.

seal6363: I think streamSession won't have any data stored to allow us to create the challenge again if the stream connection is new.
or am I missing something? I am unsure about how to accurately re-issue the challenge if there's no data.

server/service/service.go
```diff
@@ -218,6 +219,104 @@ const (
...
+	if err := stream.Send(challengeResp); err != nil {
+		return err
+	}
+	session.currentState = stateTPM20NonceSent
+	log.Infof("Sent nonce challenge to IDevID device: %s", deviceID)
+	return nil
+}
+
+// establishSessionAndSendChallenge is a helper to establish session and send challenge for TPM2.0 with IdevID.
+func (s *Service) establishSessionAndSendChallenge(stream bpb.Bootstrap_BootstrapStreamServer, ctx context.Context, lookup *types.EntityLookup, identity *bpb.Identity, ccSerial string) (*streamSession, string, error) {
```
gmacf, line 284: Instead of having this function return a new streamSession, could you update it so that it simply updates the existing one? It might save us from some complicated bugs if we just make sure that we create one session in the first connection and then update it as the stream happens.

server/service/service.go
```diff
@@ -384,7 +437,42 @@ func (s *Service) BootstrapStream(stream bpb.Bootstrap_BootstrapStreamServer) er
...
+				lu, ccSerial, err := buildLookupFromReportStatus(ctx, req.ReportStatusRequest)
+				if err != nil {
+					return err
+				}
+
+				newSession, newDeviceID, err := s.establishSessionAndSendChallenge(stream, ctx, lu, req.ReportStatusRequest.GetIdentity(), ccSerial)
+				if err != nil {
+					return err
+				}
+				newSession.pendingStatusReport = req.ReportStatusRequest
```
gmacf, line 456: What's the need for pendingStatusReport?

seal6363: sorry that was a left over code, removed

server/service/service.go
seal6363: ok, that makes sense.

gmacf (approved):

server/service/service.go
```diff
@@ -218,6 +218,103 @@ const (
...
+	if err := stream.Send(challengeResp); err != nil {
+		return err
+	}
+	session.currentState = stateTPM20NonceSent
+	log.Infof("Sent nonce challenge to IDevID device: %s", deviceID)
+	return nil
+}
+
+// establishSessionAndSendChallenge is a helper to establish session and send challenge for TPM2.0 with IdevID.
+func (s *Service) establishSessionAndSendChallenge(session *streamSession, stream bpb.Bootstrap_BootstrapStreamServer, ctx context.Context, lookup *types.EntityLookup, identity *bpb.Identity, ccSerial string) (string, error) {
```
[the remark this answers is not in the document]
seal6363: done

PR #314, opened 2026-05-09
Yyamout/public bootz server updates

youssefyamoutyy:
This PR adds the public Bootz server support needed for the first featureprofiles Bootz FNT case.

**feature profile PR**: https://github.com/openconfig/featureprofiles/pull/5016

Included in here:
- IP SAN support for generated test certificates
- support for disabling `BootstrapStream` so tests can use the unary flow
- CertZ profile plumbing in bootstrap data
- PathZ and CredZ propagation in bootstrap data

This is intended to support the initial public featureprofiles Bootz test flow and close some gaps between the current public server behaviour and the test environment.

A couple of notes:
- the first public featureprofiles test is still the minimum-config flow
- PathZ and CredZ are not exercised by that first test yet, even though this branch includes the server-side plumbing for them
- the current featureprofiles side still works around the server’s AuthZ inventory requirement for that minimum-config case

So this PR gets the public server into the right shape for the first test. More tests will follow up validation, and we will have medications here as needed!

Adding for reference: https://github.com/openconfig/featureprofiles/pull/5016

### DIFF
--- a/server/entitymanager/entitymanager.go
+++ b/server/entitymanager/entitymanager.go
@@ -36,6 +36,7 @@ import (
 	bpb "github.com/openconfig/bootz/proto/bootz"
 	epb "github.com/openconfig/bootz/server/entitymanager/proto/entity"
 	apb "github.com/openconfig/gnsi/authz"
+	ppb "github.com/openconfig/gnsi/pathz"
 )
 
 const defaultRealm = "prod"
@@ -74,6 +75,14 @@ func (m *InMemoryEntityManager) ResolveChassis(ctx context.Context, lookup *serv
 	if err != nil {
 		return nil, err
 	}
+	credzConf, err := m.populateCredentialsConfig(chassis)
+	if err != nil {
+		return nil, err
+	}
+	pathzConf, err := m.populatePathzConfig(chassis)
+	if err != nil {
+		return nil, err
+	}
 	authzConf, err := m.populateAuthzConfig(chassis)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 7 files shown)


Chounoki (approved):

Chounoki (commented):
Please wait for the review from @gmacf

server/service/service.go
Chounoki, line 1: Could you sync your branch to the latest of main branch?
You change was done on a stale version.

Chounoki (changes requested):

server/server.go
```diff
@@ -90,9 +90,37 @@ type InterceptorOpts struct {
 
 func (*InterceptorOpts) isbootzServerOpts() {}
 
+// DisableBootstrapStream disables the deprecated BootstrapStream RPC while
+// keeping unary RPCs available.
+type DisableBootstrapStream struct{}
```
Chounoki, line 95: We don't need to disable Streaming Bootz on the global sever level.
This is already implemented on a per chassis level.

https://github.com/openconfig/bootz/blob/02faee8db760811f36472c0ed14e3533f7b58c50/server/service/service.go#L693

https://github.com/openconfig/bootz/blob/02faee8db760811f36472c0ed14e3533f7b58c50/server/service/service.go#L919

Chounoki (changes requested):

server/entitymanager/entitymanager.go
Chounoki, line 1: We are going to rewrite the entire EntityManager soon to utlize the "ChassisManager" interface.
https://github.com/openconfig/bootz/blob/02faee8db760811f36472c0ed14e3533f7b58c50/server/service/service.go#L82

It probably doesn't affect your PR, only as a heads-up.

gmacf (changes requested):
Please address merge conflicts and Michael's comments, otherwise LGTM.

gmacf:
I believe this PR is now obsolete since we've refactored the service and removed EntityManager. I will take a look at re-implementing it with the new changes.

gmacf:
Actually, I believe we can close this without needing additional changes. 

https://github.com/openconfig/bootz/blob/main/testdata/bootz_config.textproto now controls all data used in Bootz, including pathz/credz policies for the device, whether streaming is supported and what certificate to use.

PR #316, opened 2026-05-12
DHCP-less Bootz spec changes

gmacf:


README.md
```diff
@@ -260,33 +260,72 @@ where the device can be interrogated from a trusted system to enroll the TPM and
...
       in this TLS handshake.
-   3. The responses from the bootz-server are signed by ownership-certificate.
+   4. The responses from the bootz-server are signed by ownership-certificate.
       The device validates the ownership-voucher, which authenticates the
       ownership-certificate. The device verifies the signature of the message
       body before accepting the message.
-   4. If the signature could not be verified, the bootstrap process starts
-      from Step 1.
+   5. If the signature could not be verified, the bootstrap process starts
+      from the respective entry point (Step 1).
```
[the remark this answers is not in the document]
gmacf: The list items are ordered correctly

Chounoki (approved):

README.md
```diff
@@ -260,11 +260,29 @@ where the device can be interrogated from a trusted system to enroll the TPM and
...
+#### DHCP (default)
+
+By default, devices are expected to perform a standard DHCP boot via the active
+control card's out-of-band (OOB) management interface. The DHCP server passes an
+option to the device for the Bootz URI from which the boot package can be
+retrieved.
+
+#### DHCP-less (inband)
+
+In environments where DHCP is not available or out-of-band (OOB) management
```
NishadCM, line 274: If device need to do dhcp-less bootz if local config for dhcp-less is available on devcie, do we need the default option on device to set as "dhcp-less" option. 

>
as per above definition the default behaviour is to use the dhcp based discovery.

- what should be the criteria for fall back to dhcp-less.
- "local configuration" (specific to dhcp-less config), is this the trigger tigger for using "dhcp-less", if config present always use "dhcp-less".
(or do the devcie need to run regular dhcp based discovery and if no success (max try), look for dhcp-less method)

Is this case(dhcp-less) applied only manually initiation, is this applied to below cases, if applied what will be the behaviour for following cases.

1) factory reset 
     > only dhcp based discovery is possible
2) reload (both can be active)
     > assumed, in this case device will look for local persistent cfg, if present perform dhcp-less
3) reimage 
     > make sure persistent config is not removed
     > assume device should, use dhcp-less if local config present
4) manual initiate
     > devcie should follow the initiate commad, if no specific option given, use dhcp-less lookup as first method.

README.md
```diff
@@ -275,9 +293,30 @@ can install production configuration and certificates into the device.
    4. The format of the DHCP message (other than response option code) follows
       [RFC](https://www.rfc-editor.org/rfc/rfc8572#page-56).
       1. The URI will be in the format of `bootz://<hostname or ip>:<port>`
+   - **Option B: DHCP-less**
+   1. A network operator manually configures the device with a local
+      configuration that gives it reachability to the Bootz server. This
+      includes static routing, interface configuration and local admin
+      credentials.
```
NishadCM, line 300: If local admin credentials are already configured.

- what will be the exit criteria for bootz work, as it is going to run in backgroud till success.
- bootz work flow involves reload/reimage and the exit criteria is the detection of configuration, If devcie is already contains configuration bootz will assume device is already provisioned, otherwise will attempt bootz starting with dhcp discovery
(with new case will be one of dhcp discovery or dhcp-less).

If we have admin creds on device before starting bootz, if bootz workflow reloads the devcie, bootz run will go in background, even if the user logs in bootz will be still running in background, need  explicit terminate from user.

README.md
```diff
@@ -275,9 +293,30 @@ can install production configuration and certificates into the device.
...
+   2. A network operator triggers DHCP-less Bootz via CLI, providing the
+      Bootz Server URI and Source Interface. These are saved to a
+      persistent parameters file on disk to survive reboots.
```
NishadCM, line 303: What will the scope and life time of this persistent data.
- If this data is present on device, every time when device gets a reload/reiame (manual), there is a risk of device starting dhcp-less workflow, as there is no other exit criteria (user cred is not an exit criteria now).

README.md
```diff
@@ -275,9 +293,30 @@ can install production configuration and certificates into the device.
...
+   3. The device enters a Bootz loop, attempting to connect to the
+      specified Bootz server from the source interface using its local
+      configuration. It should not perform a disk wipe or factory reset
+      during the initiation phase, though a reboot may be performed if
+      required.
+   4. If the DHCP-less Bootz process fails at any point, the device MUST
+      revert back to the operator-provided local configuration and attempt
+      to connect to the Bootz server again. The device MUST remain in this
+      recovery loop until either:
```
NishadCM, line 312: on failure:
- device should fall back to initial configuration.
assume that device need to retry bootz again (dhcp-less mode).

README.md
```diff
@@ -275,9 +293,30 @@ can install production configuration and certificates into the device.
...
+      1. Bootz completes successfully
+      2. The operator manually resets the device to standard DHCP mode
+         via the CLI, at which point Option A takes effect.
```
NishadCM: on this mode change(removing cfg/dhcp-less persistent config) do we need to exit current instance of bootz ?
- do we need autostart of bootz on mode chnage, 
   (or wait for user to trigger the again using manual reload/initiate command ?)

README.md
gmacf: I don't quite follow your comment so please let me know if I've misunderstood.

DHCP-less Bootz should only ever be attempted if someone has explicitly enabled it via the CLI. The implementation of this is up to the vendor but you can think of this as writing something to the filesystem that enables this feature. The existence (or lack of) startup config should have no effect on what Bootz mode is used.

To answer your questions:

> 1. Factory reset

Yes, this would go straight to DHCP mode since an operator didn't run the DHCP-less CLI after the factory reset.

> 2. reload (both can be active)

Device will look for persistent Bootz parameters and start dhcp-less Bootz if so. If they don't exist, don't enter Bootz loop.

> 3. reimage

Assuming this means an in-place OS upgrade/downgrade without touching the configuration? Again, this would depend if the Bootz parameters have been persisted on the device. If not, then it would use the existing local config without entering Bootz.

> 4. manual initiate

The CLI command is intended specifically for the DHCP-less method. For triggering DHCP Bootz, then a normal factory reset or config wipe + reload would work.

README.md
gmacf: The exit criteria is actually that the device sends a final ReportStatus gRPC to the Bootz server which indicates that it has finished bootstrapping. As I mentioned above, this would mean that the presence of a local config on the device doesn't determine whether Bootz is started or not. The persisted Bootz parameters is authoritative here.

README.md
gmacf: These persistent Bootz parameters exist for the lifetime of the Bootstrapping process. They are deleted when Bootz finishes successfully or when the operator manually cancels DHCP-less mode. This is explained further down in the doc.

README.md
gmacf: Yes, the device would keep trying dhcp-less mode on failure, but must revert to the previous local configuration (and not an empty factory-reset  configuration).

README.md
gmacf: If Bootz is successful, then you exit Bootz because it is not needed anymore.

If you manually cancel DHCP-less Bootz mode, you should stop the current Bootz attempt, and reloads into DHCP mode. See the sequence diagrams below.

README.md
NishadCM: # DHCP-less Bootz — Exit Behavior Gap

## Core Issue

**DHCP-less bootz has no exit criteria other than success.** Unlike regular bootz, the DHCP-less workflow bypasses the normal exit conditions (e.g., "exit if device is already provisioned"). Because the DHCP-less trigger is stored in persistent disk configuration, bootz will run forever in background — surviving reloads, reimages, and shutdown/boot cycles. The only way to stop it today is a factory reset or explicit user intervention (stop/remove dhcp-less config).

---

## Existing Bootz Exit Behavior (Regular / DHCP-based)

Bootz exits under the following conditions today:

| Workflow | Exit Condition |
|----------|---------------|
| Normal boot(reload/reimage) | Exits if device is already provisioned |
| bootz | On success (provisioning complete) |
| bootz | On failure — bootz does **not** exit, retries forever |
| Automated (DHCP-based) | Exits on first user creation (device is considered provisioned) |
| Manual bootz start(cli) | Exits on bootz terminate cli execution|
| Manual bootz start(cli) | Exits on reload, and do not start on boot if device is already configured |

> **Note:** In normal cases, bootz will not be running or started automatically if the device is already provisioned (through manual provisioning or prior bootz provisioning).

---

## DHCP-less Bootz

- Device already contains configuration (user/reachability config present).
- Whether to use dhcp-less or regular discovery is determined by the persistent configuration on disk.
- dhcp-less config is saved on disk, and once created will not get cleared on applying new xr-config/commit replace.

### Steps to Initiate DHCP-less Bootz

1) user already logged in to device (user configuration is present)
2) added additional configuration (reachability)
3) added dhcp-less config entry
4) manually started using bootz cli/reload

### What Happens After Reboot
- device will boot and bootz starts execution
- looks for "dhcp-less", as first steps (regular bootz workflow will exit if it detects user-name configuration)
- If "dhcp-less" is configured device will attempt "dhcp-less" bootz, and skips user config checks.

**Scenario 1: Bootz fails to reach the server**
- device is not able to complete bootz with server (could be any reason).
- bootz will continue to run in background, server will get status based on the type of failure.

**Scenario 2: User logs in while bootz is still running (after continuous failures)**
- user is logging in to device (through console/**).
- user will be prompted with an authentication window, and user will get access into the device.
- bootz will still be running in the background.
(Expecting the user to manually stop bootz or remove the dhcp-less configuration after logging in is not always feasible and is error-prone).

### Impact

1) If user is not explicitly stopping bootz, device will have bootz running in background.
2) The bootz start (dhcp-less) is associated with a persistent configuration, (if present) bootz will get started even after reboot/reimage/shutdown and boot.
3) If user attempted dhcp-less bootz, on failure then decides to provision device manually:
   - bootz will run in background even after the device is provisioned.
   - there are no exit cases for bootz other than success (which will alter the manually provisioned data).
   - bootz will be continuously attempted on device, and it will survive reload/reimage/shut and boot.
4) With stale dhcp-less data on disk, bootz will run forever in background — there is no dhcp-less workflow-specific exit criteria in automated workflow to exit bootz (except factory reset).

### Summary

DHCP-less bootz has no graceful exit path other than success or factory reset. This means:

- Stale DHCP-less config on persistent disk = bootz runs indefinitely
- User manually provisioning after a failed DHCP-less attempt = bootz still running in background
- Bootz workflow survives/gets started even after reload/reimage/shutdown

### Open Question: Should DHCP-less Bootz Require Explicit Preparation Steps?

Should the dhcp-less workflow require the following steps before initiation:

1) bootz cleanup
2) commit-replace/config removal
3) apply required configuration for reachability
4) dhcp-less command initiate (which will initiate bootz with a reload, or start without reload)

> **Note:** Here User-name configuration is still an exit criteria for bootz. On device boot, even if bootz finds a dhcp-less configuration, if the device is configured with a user configuration, bootz will decide to exit.

-> or dhcp-less work flow require a max retry and exit.
-> or start the dhcp-less with a expiry.

README.md
NishadCM: ok

README.md
gmacf: I don't think username configuration is a good exit criteria for Bootz. There are some cases where the pre-Bootz local configuration will want to have a local user configured and still have DHCP-less Bootz continue.

I do agree though that there should be some other way to prevent DHCP-less Bootz from running in the background indefinitely. Let me chat with a few of the openconfig maintainers and get back to you.

README.md
```diff
@@ -671,6 +762,43 @@ while TPM 1.2 systems are not supported.
...
+
+### DHCP-less CLI Specification
+
+Vendors supporting DHCP-less Bootz MUST implement the following standardized CLI
+commands:
+
+*   **Initiate DHCP-less Bootz**:
+
+    ```
+    bootz no-dhcp src_interface <interface> bootz_uri <bootz_uri>
```
cpeng90: When DHCP is not used, the IP configuration shall be provisioned manually, including the IP address, prefix length, and next-hop address, with support for both IPv4 and IPv6. Would you prefer to including these information in the bootz no-dhcp command?

gmacf: No, we'd prefer to keep the CLI command as simple as possible and rely on the local configuration on the device to provide reachability to the Bootz server. If we had to put all this information in the CLI command, it would become too complex due to all the differences in vendor hardware and network configurations.

README.md
gmacf: I've just pushed a new commit which addresses this. There are now two ways to stop the background Bootz process:

* Via the `bootz no-dhcp terminate` CLI, or
* By commiting a configuration on the device *after* no-dhcp mode has started.

README.md
cpeng90: Since the local network configuration is decoupled from the DHCP-less BootZ CLI, the src_interface parameter shall be removed, as the outgoing connection path is determined by the local network configuration.

gmacf: Consider the case where there are multiple interfaces/routes configures on the router before Bootz. We want to give the network operator full control over which of those interfaces gets used for Bootz. For this reason we want to keep the source interface parameter in the CLI.

README.md
```diff
@@ -275,9 +293,37 @@ can install production configuration and certificates into the device.
...
+      configuration. It should not perform a disk wipe or factory reset
+      during the initiation phase, though a reboot may be performed if
+      required.
+   4. If the DHCP-less Bootz process fails at any point, the device MUST
+      revert back to the operator-provided local configuration and attempt
+      to connect to the Bootz server again. The device MUST remain in this
+      recovery loop until either:
+      1. Bootz completes successfully
+      2. The operator manually resets the device to standard DHCP mode
+         via the CLI (`bootz no-dhcp reset`).
```
LimeHat, line 315: IMO vendor-specific cli shouldn't be part of the spec

README.md
```diff
@@ -275,9 +293,37 @@ can install production configuration and certificates into the device.
...
+      3. The operator explicitly terminates the Bootz process manually
+         via the CLI (`bootz no-dhcp terminate`).
```
LimeHat, line 317: IMO vendor-specific cli shouldn't be part of the spec

README.md
```diff
@@ -275,9 +293,37 @@ can install production configuration and certificates into the device.
...
+      4. A manual configuration commit is completed on the device. The
```
LimeHat: Is this required? In my view, this unnecessarily complicates the state machine, and the semantics are unclear:
- Some implementations may not allow commits when the device is in the ZTP/bootz mode, or
- Not all configuration interfaces necessarily have the notion of commits.

I'd suggest dropping this requirement.

README.md
```diff
@@ -671,6 +782,59 @@ while TPM 1.2 systems are not supported.
...
+  ```
+
+  - This command stops the DHCP-less Bootz loop.
+  - It MUST immediately kill the running background Bootz agent.
+  - It MUST delete the persistent Bootz parameter file from the disk.
+  - It MUST NOT wipe or alter the current local configuration (running or startup).
+  - It transitions the device out of bootstrap state, leaving all current
+    routing, interface, and credential configurations intact for local operations.
+
+- **Logging**: The device MUST log initiation, exit, and reset events to
```
LimeHat: The earlier paraph talks about returning the device to the factory default state after issuing the `reset` command, which means all local logs must be wiped. Unless the operator adds a remote syslog server, I don't see how the reset can be logged.

README.md
```diff
@@ -671,6 +782,59 @@ while TPM 1.2 systems are not supported.
...
+
+- **Exit/Reset DHCP-less Bootz**:
+
+  ```bash
+  bootz no-dhcp reset
+  ```
+
+  - This command stops the DHCP-less Bootz loop.
+  - This is a destructive operation. It MUST wipe the temporary
+    pre-configuration (restoring the device back to its factory-default state)
```
LimeHat: Wipe the config, or restore the factory-default state? These are two different things (the latter implies much broader cleanup that removes all operator data, including logs, temporary files, other non-config files, etc)

README.md
```diff
@@ -260,11 +260,29 @@ where the device can be interrogated from a trusted system to enroll the TPM and
...
+### Operating modes
+
+#### DHCP (default)
+
+By default, devices are expected to perform a standard DHCP boot via the active
+control card's out-of-band (OOB) management interface. The DHCP server passes an
+option to the device for the Bootz URI from which the boot package can be
+retrieved.
+
+#### DHCP-less (inband)
```
LimeHat: Generally speaking, "DHCP-less" and "inband" and not synonymous.  
For example, there are plenty of implementation that can do DHCP-based ZTP/sZTP over inband ports (and this logic can be extended to bootz if needed). Additionally, nothing prevents from using this dhcp-less method with the OOB interface.

I'd suggest removing references to the `inband`.

LimeHat (changes requested):

README.md
cpeng90: Is interface parameter optional or mandatory?

README.md
```diff
@@ -671,6 +762,43 @@ while TPM 1.2 systems are not supported.
      will finally send out an empty `ReportStatusResponse` message to
      acknowledge the status report. If the challenge fails, an error will be
      returned and the device must start over from Step 7.
+9.   Cleanup
+   - If the device was bootstrapped via DHCP-less Bootz, it MUST
+     now automatically delete the persistent Bootz parameter file and wipe
+     the temporary pre-configuration used for initial connectivity.
```
cpeng90: Does it imply that the device must receive initial config at step 6.ix? 

- If so, the cleanup needs to be done when the initial config is applied. 
- if there is no initial config is received at step 6.ix, the cleanup must be done when full configuration is applied.

In either case, the step 9 seems not in the right place.

README.md
cpeng90, line 707: "... with the intent for each of the device's control cards..." It is platform-dependent, and in many active-standby designs the standby is not independently provisioned.

README.md
```diff
@@ -740,6 +904,187 @@ This is the preferred workflow for security considerations. This workflow
...
+    Server-->>Device: Error / Timeout
+    deactivate Server
+    Device->>Device: Revert to pre-Bootz config
+    Note over Device: Retry loop continues indefinitely
+
+    Operator->>Device: Commit new startup config manually
+    Device->>Device: Wipe Bootz parameters from file system
+    Device->>Device: Exit Bootz loop
+
+    deactivate Device
```
NishadCM, line 1054: during bootz we have 3 commits that can happen (commit id will chnage).
1) Revert to pre-Bootz config
2) bootconfig/oc config (immutable)
3) dynamic cfg (bootconfig/oc)
these commits also will update config, bootz will not be able to differentiate between the actual commit trigger.
eg: item 2 above, boot-config is loaded as a startup config, same as manual startup config load.

README.md
gmacf: I'd prefer to keep this mandatory, that way there is no possibility of ambiguity when selecting the interface.

README.md
gmacf: We don't want the cleanup to happen immediiately after an initial config is received and committed during step 6. This is because a successful commit doesn't necessarily mean that the device has reachability. Only after the device successfully sends a bootstrap status success, do we know that the config is good and it can wipe the pre-configuration.

Basically, this needs to be the absolute last step so that we can recover in the event of any failures found between the initial commit received and the final report.

README.md
gmacf: I have dropped the commit interception requirement as it was too complex

README.md
gmacf: Done, thanks.

README.md
gmacf: Clarified here that it should not perform a factory reset/disk wipe, and only wipe the preconfiguration and Bootz parameters.

README.md
gmacf: Clarified that this does not perform a full disk wipe

README.md
gmacf: Good point, I have dropped this requirement.

README.md
gmacf: I have reworded this section to have recommendations for the CLI format but not enforced. From our perspective, we want this command to be standardized where possible to reduce automation complexity when dealing with multiple vendors.

README.md
gmacf: If the device reports itself as a chassis with two supervisors, Bootz will provide a desired state for both. Since the active supervisor gets both of these, it has full control over how or whether it distributes the desired state to the standby. We don't enforce this in the spec, which allows for some flexibility between platforms.

gmacf:
FYI to reviewers: I'll be merging this tomorrow unless there are further comments

NishadCM (approved):

PR #320, opened 2026-05-18
Add server/config package.

Chounoki:


server/proto/BUILD.bazel
```diff
@@ -0,0 +1,53 @@
...
+    ],
+)
+
+##############################################################################
+# Go
+##############################################################################
+
+go_proto_library(
+    name = "config_go_proto",
+    importpath = "github.com/openconfig/bootz/server/proto",
```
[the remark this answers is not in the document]
Chounoki: Done.

server/proto/config.proto
```diff
@@ -0,0 +1,85 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package config;
```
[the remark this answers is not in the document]
Chounoki: Done.

server/proto/config.proto
```diff
@@ -0,0 +1,85 @@
...
+  // Bootz server address (IP:port).
+  string server_address = 1;
+  // Bootz server trust anchor cert key pair.
+  CertKeyPair trust_anchor = 2;
+  // Owner certificate key pair.
+  CertKeyPair owner_certificate = 3;
+  // Based64 encoding of ASN.1 DER vendor CA certificates.
+  repeated string vendor_ca_certs = 4;
+  // Chassis owned by the orgization.
+  repeated ChassisInventory chassis_inventories = 5;
```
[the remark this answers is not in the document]
Chounoki: This proto will be mostly used in the format of `textproto` as data source, and must allow easy editing by humans when needed.

The representation of `bytes` in `textproto` is really ugly (e.g. "\x01\x02\x03\x04...") and difficult to edit/replace manually, so it is better to keep based64 encoding even though it adds an extra layer of conversion.

server/proto/config.proto
```diff
@@ -0,0 +1,85 @@
...
+}
+
+message ControlCard {
+  // Serial number of the control card.
+  string serial_number = 1;
+  // Base64 encoding of ownership voucher.
+  string ownership_voucher = 2;
+  // Base64 encoding of PKIX DER EK/PPK public key.
+  // This field is not populated for control cards having IDevID.
+  string public_key = 3;
```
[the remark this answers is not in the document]
Chounoki: Same reason as above.

server/proto/config/config.pb.go
```diff
@@ -0,0 +1,461 @@
+// Code generated by protoc-gen-go. DO NOT EDIT.
+// versions:
+// 	protoc-gen-go v1.36.11
+// 	protoc        v7.34.1
+// source: github.com/openconfig/bootz/server/proto/config.proto
+
+package proto
```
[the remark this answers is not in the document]
Chounoki: Done.

server/proto/config.proto
```diff
@@ -0,0 +1,85 @@
...
+  repeated string vendor_ca_certs = 4;
+  // Chassis owned by the orgization.
+  repeated ChassisInventory chassis_inventories = 5;
+}
+
+message CertKeyPair {
+  // Base64 encoding of ASN.1 DER certificate.
+  string cert = 1;
+  // Base64 encoding of PKCS#8 DER private key.
+  string key = 2;
```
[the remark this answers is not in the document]
Chounoki: This proto will be mostly used in the format of textproto as data source, and must allow easy editing by humans when needed.

The representation of bytes in textproto is really ugly (e.g. "\x01\x02\x03\x04...") and difficult to edit/replace manually, so it is better to keep based64 encoding even though it adds an extra layer of conversion.

gmacf:
Is this for Monax? If so, I would like to see how this fits in with the existing protos or proposed workflow. Do you have any example configs for this?

Chounoki:
> Is this for Monax? If so, I would like to see how this fits in with the existing protos or proposed workflow. Do you have any example configs for this?

Yes, this is for Monax, so the vendors can use all real data like ownership_voucher.
I haven't created the PR yet, but when creating the Bootz server, this `Config` proto should be read from a file and provided to the Bootz server.

server/proto/config.proto
```diff
@@ -0,0 +1,87 @@
...
+import "github.com/openconfig/bootz/proto/bootz.proto";
+import "github.com/openconfig/gnsi/authz/authz.proto";
+import "github.com/openconfig/gnsi/pathz/pathz.proto";
+
+option go_package = "github.com/openconfig/bootz/server/proto/config";
+
+// A binding configuration for Bootz server.
+message Config {
+  // Bootz server address (IP:port).
+  string server_address = 1;
```
gmacf, line 29: How will we know the server address of the Bootz server ahead of time? Some runtime environments (e.g. Google's) don't let you reserve a specific IP when bringing up SUT components. We will need another way to learn about the Bootz server address.

server/proto/config.proto
```diff
@@ -0,0 +1,87 @@
...
+}
+
+message CertKeyPair {
+  // Base64 encoding of ASN.1 DER certificate.
+  string cert = 1;
+  // Base64 encoding of PKCS#8 DER private key.
+  string key = 2;
+}
+
+message ChassisInventory {
```
gmacf: Could this just be 'Chassis'? Inventory sounds like a collection rather than a single device.

server/proto/config.proto
```diff
@@ -0,0 +1,87 @@
...
+  // cards.
+  repeated ControlCard control_cards = 2;
+  // The intended hostname of the chassis.
+  string hostname = 3;
+  // Boot mode defines the boot mode that can be secure or insecure.
+  bootz.BootMode boot_mode = 4;
+  // Whether Streaming Bootz is supported or not.
+  bool streaming_supported = 5;
+  // Software image to be loaded on the chassis.
+  bootz.SoftwareImage intended_image = 6;
```
gmacf, line 61: The Bootz SoftwareImage contains a download URL. How will we know that URL ahead of time? This URL should be generated by the ImageService SUT dynamically.

gmacf:
> > Is this for Monax? If so, I would like to see how this fits in with the existing protos or proposed workflow. Do you have any example configs for this?
> 
> Yes, this is for Monax, so the vendors can use all real data like ownership_voucher. I haven't created the PR yet, but when creating the Bootz server, this `Config` proto should be read from a file and provided to the Bootz server.

In that case, could you either edit the existing Monax protos under server/tests/proto or move the config there? The reason I ask is because this file is very similar to server/tests/proto/test.proto

server/proto/config.proto
Chounoki: Here is my understanding:

With Monax, each SUT will run in a container, which by default is assigned to a private network by Docker runtime. This private network is bridged (NAT'ed) by the host machine. (I.e. the host machine acts as the gateway of the SUT's private network).

https://docs.docker.com/engine/network/

This private network address doesn't really matter, because any external switch will only see the real IP address of the host machine on which the containers run. In other words, the host machine does port mapping from container private network to host public network.

The DHCP entries must use the real host IP address as Bootz server address. Similarly the image URI created by Bootz must also use the real host IP address, because to any external switches, only the real IP address of the host machine is visible (All IP addresses inside the private network of the containners are not visible externally).

server/proto/config.proto
Chounoki: Sure, I will change to that.

server/proto/config.proto
Chounoki: I think the image URI must also use the real IP address of the host machine, which is known by the user doing the test beforehand.

In the worst case, Bootz can still override this IP if required (i.e. we can provide a function for Monax to call).

Chounoki:
> > > Is this for Monax? If so, I would like to see how this fits in with the existing protos or proposed workflow. Do you have any example configs for this?
> > 
> > 
> > Yes, this is for Monax, so the vendors can use all real data like ownership_voucher. I haven't created the PR yet, but when creating the Bootz server, this `Config` proto should be read from a file and provided to the Bootz server.
> 
> In that case, could you either edit the existing Monax protos under server/tests/proto or move the config there? The reason I ask is because this file is very similar to server/tests/proto/test.proto

Besides Monax, this proto definition will be referenced by `artifact_manager` and `chassis_manager` as argument in their respective New() function, as well as by `server_emulator` and `client_emulator`, so I am not sure whether putting it into the `test` directory is appropriate.

Even without doing any test (Monax or not), this proto definition is still required.

server/proto/config.proto
gmacf: Monax isn't necessarily going to run on Docker containers. Some vendors may run these SUT components on bare-metal machines, and as I said above, it definitely wouldn't work in Borg. This is why we should find a solution that can be applied to any runtime/deployment environment.

gmacf:
> > > > Is this for Monax? If so, I would like to see how this fits in with the existing protos or proposed workflow. Do you have any example configs for this?
> > > 
> > > 
> > > Yes, this is for Monax, so the vendors can use all real data like ownership_voucher. I haven't created the PR yet, but when creating the Bootz server, this `Config` proto should be read from a file and provided to the Bootz server.
> > 
> > 
> > In that case, could you either edit the existing Monax protos under server/tests/proto or move the config there? The reason I ask is because this file is very similar to server/tests/proto/test.proto
> 
> Besides Monax, this proto definition will be referenced by `artifact_manager` and `chassis_manager` as argument in their respective New() function, as well as by `server_emulator` and `client_emulator`, so I am not sure whether putting it into the `test` directory is appropriate.
> 
> Even without doing any test (Monax or not), this proto definition is still required.

I'm not sure how this config proto will be used in the production (non-test instance). Doesn't this config have to contain an exhaustive list of the inventory and other artifacts? In the real world, we fetch these from external systems.

server/proto/config.proto
Chounoki: If the vendors are going to directly run the Monax test we provide to them, then each SUT will be inside a container just like the [math example provided by Monax](https://github.com/openconfig/monax/tree/main/example/math), you can see each SUT has its own [Dockerfile](https://github.com/openconfig/monax/blob/main/example/math/addition/deploy/Dockerfile) and [kubernetes.txtpb](https://github.com/openconfig/monax/blob/main/example/math/addition/deploy/kubernetes.txtpb) file.

If the vendors don't plan to run the Monax, but to use each service on bare machine without Monax, then they already know the IP of the machine which can be manually written into the textproto of this file by the vendors, and then Bootz server will be initialized from this textproto.

In either case, personally I think it is better and safer to ask the vendors to manually fill in the IP address of the machine on which they plan to run the test (regardless of Bootz server inside a container or not), since the machine can have multiple network interfaces and multiple IP addresses.

While Bootz calls `net.Listen()` function without specifying an IP, it will create a listener on each of those network interfaces and their IP addresses (https://pkg.go.dev/net#Listen), which means if we try to retrieve a single server IP address dynamically from Bootz runtime, we would get multiple IP addresses instead, and it would be a mess because we can only put one single Bootz IP into DHCP.

Chounoki:
> > > > > Is this for Monax? If so, I would like to see how this fits in with the existing protos or proposed workflow. Do you have any example configs for this?
> > > > 
> > > > 
> > > > Yes, this is for Monax, so the vendors can use all real data like ownership_voucher. I haven't created the PR yet, but when creating the Bootz server, this `Config` proto should be read from a file and provided to the Bootz server.
> > > 
> > > 
> > > In that case, could you either edit the existing Monax protos under server/tests/proto or move the config there? The reason I ask is because this file is very similar to server/tests/proto/test.proto
> > 
> > 
> > Besides Monax, this proto definition will be referenced by `artifact_manager` and `chassis_manager` as argument in their respective New() function, as well as by `server_emulator` and `client_emulator`, so I am not sure whether putting it into the `test` directory is appropriate.
> > Even without doing any test (Monax or not), this proto definition is still required.
> 
> I'm not sure how this config proto will be used in the production (non-test instance). Doesn't this config have to contain an exhaustive list of the inventory and other artifacts? In the real world, we fetch these from external systems.

This config profo is only for the open source Bootz. This config proto itself as well as its consumers: `artifact_manager`, `chassis_manager`, `bootz_emulator` and `client_emulator` won't be sync'ed back to our internal codebase.

Our internal Bootz will use a different version of `artifact_manager` and `chassis_manager`, which implements the same interfaces as the open source ones.

Chounoki:
> > > > > > Is this for Monax? If so, I would like to see how this fits in with the existing protos or proposed workflow. Do you have any example configs for this?
> > > > > 
> > > > > 
> > > > > Yes, this is for Monax, so the vendors can use all real data like ownership_voucher. I haven't created the PR yet, but when creating the Bootz server, this `Config` proto should be read from a file and provided to the Bootz server.
> > > > 
> > > > 
> > > > In that case, could you either edit the existing Monax protos under server/tests/proto or move the config there? The reason I ask is because this file is very similar to server/tests/proto/test.proto
> > > 
> > > 
> > > Besides Monax, this proto definition will be referenced by `artifact_manager` and `chassis_manager` as argument in their respective New() function, as well as by `server_emulator` and `client_emulator`, so I am not sure whether putting it into the `test` directory is appropriate.
> > > Even without doing any test (Monax or not), this proto definition is still required.
> > 
> > 
> > I'm not sure how this config proto will be used in the production (non-test instance). Doesn't this config have to contain an exhaustive list of the inventory and other artifacts? In the real world, we fetch these from external systems.
> 
> This config profo is only for the open source Bootz. This config proto itself as well as its consumers: `artifact_manager`, `chassis_manager`, `bootz_emulator` and `client_emulator` won't be sync'ed back to our internal codebase.
> 
> Our internal Bootz will use a different version of `artifact_manager` and `chassis_manager` that implement the same interface.

I have created [another PR of `artifact_manager`](https://github.com/openconfig/bootz/pull/323) to demonstrate how this `config proto` is consumed, but due to the dependency, I have to create the new PR containing both `artifact_manager` package and this `config proto` package together in the same PR, otherwise the new PR will fail the building test due to dependency.

server/proto/config.proto
gmacf: I think my main point still stands: How will we run this Monax test on Borg if you have to know the IP address ahead of time? We need to think about all future cases, not just the immediate ones.

Monax at its core is a SUT manager. It's purpose is to automatically bring up, manage and provide access to SUT components without having to know the implementation/addressing details about them. It is responsible for figuring out the IP addresses of the SUT components it brings up, and then initiating the gRPC/HTTP/whatever connections to them. By introducing external fields that override this, you are potentially breaking that binding.

Regarding net.Listen() listening on all available IP addresses, I don't see that as a huge problem. Sure, the server will need to pick one address to provide in the DHCP lease but the beauty of Monax is that each vendor can customize that logic as much as they want. This is why I've put in a GetBootzURL() RPC to the Bootz controller: Each test implementation has full control over how the Bootz server determines what its correct address is.

Chounoki:
I think we can put Monax aside for now, because we are still too far away from implementing any Monax test.

The major use of this `config proto definition` is to replace the current `Bootz server emulator` command line arguments defined as below:
https://github.com/openconfig/bootz/blob/main/server/emulator/main.go#L38-L42

I will remove all those command line arguments and only use a single textproto file generated from this `config proto definition` as the command line argument for starting a `Bootz server emulator`.

The `client emulator` will do the same, by also using the same single textproto file generated from this `config proto definition` as the command line argument for starting a `Bootz client emulator`.

How we should start the Monax Bootz test can be discussed later and doesn't have to use this `config proto`, but we still need this proto for the server+client emulator. Currently those emulators generate lots of fake certs/keys/vouchers internally, but we can't do that anymore as we are going to ask the vendors to test with real switches, so all those things will be provided by this textproto.

I hope my explanation makes things a little clearer now.

gmacf (approved):

PR #321, opened 2026-05-19
Docker file for dhcp and http services to be run from monax.

thesrinath:


Chounoki:
Since these files are exclusively for monax, maybe we can create a root directory like "//monax" or "//test", and just like the directory structure shown in the monax example, we can put them under "//monax/deploy" or "//test/deploy".

What is your opinion?

marcushines (approved):

PR #337, opened 2026-06-10
Add Bootz test and scripts.

Chounoki:


test/run_bootz_with_http.sh
gmacf, line 1: Why do we need these scripts that spin up the emulator? Doesn't Monax handle that automatically (from bootz_test.go?)

test/monax_bootz.go
```diff
@@ -0,0 +1,78 @@
...
+	defer func() {
+		if err := sut.Stop(ctx); err != nil {
+			log.ErrorContextf(ctx, "Failed to stop SUT: %v", err)
+		}
+	}()
+
+	fmt.Println("=========================================================================")
+	fmt.Println("The SUTs (Bootz, HTTP) are now ready and running in Monax containers.")
+	fmt.Println("=========================================================================")
+	fmt.Println(">>> To test DHCP Bootz: <<<")
```
gmacf, line 64: Do we really need these? Ondatra bindings let you run gNOI and/or CLI commands on the DUT without needing any human intervention, which is going to be important for having an automated test.

test/bootz_test.go
```diff
@@ -0,0 +1,78 @@
...
+	"fmt"
+	"os"
+	"os/signal"
+	"testing"
+
+	log "github.com/golang/glog"
+	"github.com/openconfig/monax"
+	"github.com/openconfig/monax/monaxtest"
+	"github.com/openconfig/monax/runtime/kubernetesruntime"
+)
```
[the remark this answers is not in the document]
Chounoki: Done.

test/bootz_test.go
```diff
@@ -0,0 +1,78 @@
...
+	fmt.Println(">>> To test DHCP-less Bootz: <<<")
+	fmt.Println("1. Run the DHCP-less Bootz commands on your chassis to start the testing.")
+	fmt.Println("=========================================================================")
+	fmt.Println("After you finish the testing, press Ctrl+C to stop the SUTs.")
+	fmt.Println("=========================================================================")
+
+	sigchan := make(chan os.Signal, 1)
+	signal.Notify(sigchan, os.Interrupt)
+	<-sigchan
+}
```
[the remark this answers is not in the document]
Chounoki: Done.

test/run_bootz_with_http.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+set -e
+go build -o ../server/emulator/main ../server/emulator/main.go
+../server/emulator/main --alsologtostderr --config_file=./config/bootz_config.textproto --http_address=:8080 --http_folder=./www/
```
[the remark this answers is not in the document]
Chounoki: Done.

test/run_dhcp_only.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+set -e
+go build -o ../dhcp/main/dhcp ../dhcp/main/dhcp.go
+sudo -s ../dhcp/main/dhcp --alsologtostderr --dhcp_file=./config/dhcp_config.textproto
```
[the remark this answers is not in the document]
Chounoki: Done.

test/monax_bootz.go
Chounoki: We don't know how vendors are going to use Ondatra and how vendors are going to connect the hardware.
As far as I know, vendors haven't been onboarded with Ondatra yet.

For now, we let the vendors connect their switch directly to the PC to start the testing without needing anything else.
We can update this later when vendors have Ondatra set up.

test/run_bootz_with_http.sh
Chounoki: Monax can't handle DHCP service which needs to bind to the Ethernet interface on layer 2 data link to work.

Monax uses KIND (Kubernetes IN Docker) which goes over 2 layers of NAT to forward traffic to the SUTs inside. NAT is layer 3 protocol.
DHCP service sitting inside Monax will never be able to get any layer 2 traffic.

So even if we use Monax, DHCP service will always have to be created outside of Monax directly on the host machine. Because of this, there is really no merit using Monax since it is now only adding burden instead of simplifying setup.

Anyway, we still provide 2 methods to test Bootz, and vendors can choose the method most convenient to them:
1. All services (Bootz, HTTP, DHCP) directly running on the host machine by executing a single bash script.
2. Bootz+HTTP running inside Monax behind 2 layers of NAT, while DHCP directly running on the host machine using bash script.

I will create another PR containing the README.

test/run_bootz_with_http_and_dhcp.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+set -e
+go build -o ../server/emulator/main ../server/emulator/main.go
+sudo -s ../server/emulator/main --alsologtostderr --config_file=./config/bootz_config.textproto --http_address=:8080 --http_folder=./www/ --dhcp_file=./config/dhcp_config.textproto
```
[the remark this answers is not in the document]
Chounoki: Done.

test/BUILD.bazel
```diff
@@ -0,0 +1,12 @@
...
+go_test(
+    name = "bootz_test",
+    srcs = ["bootz_test.go"],
+    deps = [
+        "@com_github_golang_glog//:glog",
+        "@com_github_openconfig_monax//:monax",
+        "@com_github_openconfig_monax//monaxtest",
+        "@com_github_openconfig_monax//runtime/kubernetesruntime",
+    ],
+)
```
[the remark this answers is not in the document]
Chounoki: Done.

test/monax_bootz.go
gmacf: Every WBB vendor should be onboarded with Ondatra. It is basically a prerequisite for every single FNT in featureprofiles and has good maturity.

test/run_bootz_with_http.sh
gmacf: > Because of this, there is really no merit using Monax

Monax support is the ultimate goal for the Bootz integration test. We specifically don't want to have custom bash scripts that completely bypass the benefit of Monax. If there is a blocker (in this case DHCP service support), we should work with the Monax folks to find solutions instead of hacking together bash scripts that work in a specific scenario.

I understand this solution may get us some quick results right now, but it will cause more headaches in the future if we don't address it now.

test/monax_bootz.go
Chounoki: It is OK, this isn't necessarily the final code, we can always add Ondatra for more complicated setup later.
For Ondatra to work, vendors must first setup Ondatra and ATE with their own configurations and test vectors which will take significant amount of time.

Since the setup environment of each vendor would be different, we don't control how they connect the hardware, now we just want to give them a test requiring the least setup so that the vendors can first try whether this could work.

gmacf (approved):

Chounoki:
Thanks Gareth. We will have a discussion with Monax team for this problem.
The potential solution is actually very simple, Monax needs to remove KIND, so that all layer 2 traffic can work.

PR #354, opened 2026-07-01
Add the definiton of vendor-specific function "dut.StartBootz()".

Chounoki:


test/dut/dut.go
```diff
@@ -0,0 +1,40 @@
...
+func StartBootz(dhcp bool) error {
+	var err error
+
+	log.Infof("=============================================================================")
+	log.Infof("Starting the Bootz process on the DUT (DHCP: %v)", dhcp)
+	log.Infof("=============================================================================")
+
+	// Implement the logic here to start the Bootz process on the DUT.
+	// You can either use your own organization's private libraries to directly control your switch chassis via any vendor-specific APIs,
+	// or employ the Ondatra library (https://github.com/openconfig/ondatra) to indirecly control your switch chassis via the connected ATE if any.
```
gmacf: If we plan on giving vendors the option to use Ondatra, you need some way to pass the DUT binding to this package from the test. 

At minimum, you need to pass in a testing.TB object from the parent test. See https://pkg.go.dev/github.com/openconfig/ondatra#DUTs

test/README.md
```diff
@@ -17,19 +17,23 @@ chassis of your choice.
 
 We provide two methods to run the Bootz test.
 
-- Bare Metal Method
+- Bare Metal Manul Test Method
```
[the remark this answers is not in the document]
Chounoki: Done.

test/README.md
```diff
@@ -117,35 +128,35 @@ You can choose to test the DHCP Bootz flow or the DHCP-less Bootz flow.
...
    service.
 
    `./run_dhcp_only.sh`
 
    NOTE: DHCP service needs root privilege to bind to the Ethernet interface.
 
-1. Connect the management port of your switch chassis to the Ethernet port of
-   the PC.
+3. Run the command below in another terminal to build the Monax SUTs
+   (Bootz, HTTP), then `dut.StartBootz()` fucntion will be called to start the
```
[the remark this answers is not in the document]
Chounoki: Done.

test/dut/dut.go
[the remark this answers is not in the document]
Chounoki: Done.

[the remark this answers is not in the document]
Chounoki: Done.

test/README.md
```diff
@@ -117,35 +128,35 @@ You can choose to test the DHCP Bootz flow or the DHCP-less Bootz flow.
...
+1. Configure the management interface of your switch chassis with a static IP
    from the same subnet of the PC.
 
-3. Connect the management port of your switch chassis to the Ethernet port of
+2. Connect the management port of your switch chassis to the Ethernet port of
    the PC.
 
-4. Run the DHCP-less Bootz commands on your switch chassis to begin the testing.
+3. Run the command below in a terminal to build the Monax SUTs (Bootz, HTTP),
+   then `dut.StartBootz()` fucntion will be called to start the testing
```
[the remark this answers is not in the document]
Chounoki: Done.

test/dut/dut.go
Chounoki: My current plan is to let the vendors do all the Ondatra instantiation and bindings (if needed) inside this `dut` package.
Will that suffice?

gmacf: Generally, Ondatra bindings are acquired through the Go testing framework which is why most of those functions require access to the TB object. 

This seems to be impossible in this case because monax_bootz.go is a standalone Go binary and not a go_test and so doesn't have access to that object. Would it be possible to make this into a Go test instead of a binary?

Chounoki: Yes, I plan to change it from current binary to test later once we have a method implemented to judge the test result.

Should I change it now? Currently we don't have a way to judge and end the test with success/failure.

gmacf: I think that's fine then, let's do those changes in a future PR. For now just a TODO would be good.

gmacf (approved):

test/dut/dut.go
Chounoki: Roger. Thanks!

PR #355, opened 2026-07-06
Implement the test service.

Chounoki:


server/chassismanager/chassismanager.go
```diff
@@ -33,6 +36,8 @@ type InMemoryChassisManager struct {
 	chassis map[string]*cpb.Chassis
 }
 
+var channel chan *bpb.ReportStatusRequest
```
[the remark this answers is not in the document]
Chounoki: Done.

server/chassismanager/chassismanager.go
```diff
@@ -77,11 +82,39 @@ func (m *InMemoryChassisManager) UpdateStatus(ctx context.Context, req *bpb.Repo
...
+	if channel != nil {
+		switch req.GetStatus() {
+		case bpb.ReportStatusRequest_BOOTSTRAP_STATUS_SUCCESS, bpb.ReportStatusRequest_BOOTSTRAP_STATUS_FAILURE:
+			select {
+			case channel <- req:
+			default:
+				log.Warningf("Channel is full, dropping status update: %v", req.GetStatus())
+			}
+		}
+	}
```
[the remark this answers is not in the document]
Chounoki: Done.

server/chassismanager/chassismanager.go
```diff
@@ -77,11 +82,39 @@ func (m *InMemoryChassisManager) UpdateStatus(ctx context.Context, req *bpb.Repo
...
+		return nil, status.Errorf(codes.DeadlineExceeded, "context deadline exceeded")
+	}
+}
+
 // New returns a new in-memory chassis manager.
 func New(config *cpb.Config) *InMemoryChassisManager {
+	// If test mode is enabled, we create a channel for fetching status.
+	if config.GetTestMode() {
+		channel = make(chan *bpb.ReportStatusRequest, 1)
+	}
```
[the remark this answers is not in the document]
Chounoki: Done.

server/chassismanager/chassismanager.go
```diff
@@ -77,11 +82,39 @@ func (m *InMemoryChassisManager) UpdateStatus(ctx context.Context, req *bpb.Repo
...
+func (m *InMemoryChassisManager) FetchStatus(ctx context.Context, req *tpb.FetchStatusRequest) (*bpb.ReportStatusRequest, error) {
+	if channel == nil {
+		return nil, status.Errorf(codes.FailedPrecondition, "fetching status is only supported in test mode")
+	}
+	select {
+	case status := <-channel:
+		return status, nil
+	case <-ctx.Done():
+		return nil, status.Errorf(codes.DeadlineExceeded, "context deadline exceeded")
+	}
```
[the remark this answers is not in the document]
Chounoki: Done.

server/chassismanager/chassismanager.go
```diff
@@ -23,14 +23,19 @@ import (
...
 	bpb "github.com/openconfig/bootz/proto/bootz"
+	tpb "github.com/openconfig/bootz/proto/test"
 	cpb "github.com/openconfig/bootz/server/proto/config"
 )
 
 // InMemoryChassisManager provides a simple in memory handler for chassis.
 type InMemoryChassisManager struct {
 	chassis map[string]*cpb.Chassis
+	// channel is used to fetch bootstrap status when in test mode.
+	channel chan *bpb.ReportStatusRequest
```
gmacf: We should name this something more descriptive.

proto/test.proto
```diff
@@ -0,0 +1,34 @@
...
+import "google/protobuf/struct.proto";
+
+import "github.com/openconfig/bootz/proto/bootz.proto";
+
+option go_package = "github.com/openconfig/bootz/proto/test";
+
+// The test service definition.
+service Test {
+  // This RPC fetches the bootstrap result from the Bootz server.
+  // It blocks until the bootstrap process succeeds or fails.
```
gmacf: I wouldn't recommend using a blocking unary RPC. This is perfectly suited for a streaming RPC instead.

gmacf: See https://github.com/openconfig/bootz/blob/8290c185f6101c3ad3a55c131324e3cbc4825e4c/server/tests/proto/sut.proto#L121 for an example of what this looks like.

server/server.go
```diff
@@ -114,6 +115,10 @@ func NewServer(config *cpb.Config, opts ...Opts) (*Server, error) {
 	}
 	s := grpc.NewServer(grpc.Creds(credentials.NewTLS(conf)))
 	bpb.RegisterBootstrapServer(s, c)
+	// Register the test service only if we are in test mode.
+	if config.GetTestMode() {
+		tpb.RegisterTestServer(s, cm)
```
gmacf, line 118: You cannot use the gRPC credentials created for the device-facing gRPC service here. These credentials load the vendor trust bundle and potentially check for IDevID certs. In reality, the Monax <-> Bootz SUT gRPC communication will not use this.

You are better off creating another gRPC server in parallel.

server/server.go
```diff
@@ -114,6 +115,10 @@ func NewServer(config *cpb.Config, opts ...Opts) (*Server, error) {
 	}
 	s := grpc.NewServer(grpc.Creds(credentials.NewTLS(conf)))
 	bpb.RegisterBootstrapServer(s, c)
+	// Register the test service only if we are in test mode.
+	if config.GetTestMode() {
```
gmacf, line 117: I'm not sure if we need this check. The implementation here is going to be used for testing anyway and so it's probably fine to always start the other gRPC service.

server/chassismanager/chassismanager.go
Chounoki: Sure. Any suggestion of the name?

server/server.go
Chounoki: Since we create the RPC server using TLS config `VerifyClientCertIfGiven`, I think when we make the test RPC call in Monax test, we don't have to provide any cert and the connection will be accepted. So I guess we don't have to create another gRPC server in parallel.

https://github.com/openconfig/bootz/blob/8290c185f6101c3ad3a55c131324e3cbc4825e4c/common/tls/tls.go#L135

proto/test.proto
Chounoki: Since we always terminate the Monax test on the very first valid status report, I thought streaming RPC might be redundant, which will make both the caller code and the callee code more complex than unary RPC.

But if you prefer streaming RPC, I can change to it.

server/server.go
Chounoki: My thinking was that since this repository is open source, anyone may use this repository to build their own Bootz server for their organization's internal production bootstrapping, so it would be better not to expose this test RPC when it is not configured as a test server.

Because this test RPC does not require any credential and will become a vulnerability if exposed in production environment.

server/chassismanager/chassismanager.go
Chounoki: Done. Changed to `subscriberChannel`.

proto/test.proto
Chounoki: Done.

server/server.go
gmacf: You don't have the ability to control what authentication method Monax uses to connect to SUTs. See https://github.com/openconfig/monax/blob/0bf89ed4a9a19192824c7120c11451b80cc5a72a/grpc.go#L92

Currently it uses `insecure` which disables all transport security. Bootz server is specifically listening for TLS handshakes.

gmacf (changes requested):

server/server.go
Chounoki: Thanks for pointing this out.

This does mean we need to create a new RPC server listening on a new port.

I guess in the meanwhile we should probably also create a bug to Monax team for TLS support in future.

Chounoki: Monax team has updated their codebase with RPC call options to support secured RPC calls.

https://github.com/openconfig/monax/blob/883500dc841b87c539c575068dd7d3d6423e311b/grpc.go#L94

We don't need to create yet another insecure RPC server in parallel now.

gmacf (approved):

PR #358, opened 2026-07-15
Add ipv6 support to bootz server.

chaskayamane:
The ipv6 format uses colons, `:`, in the address so splitting on colon to find port does not work for ipv6 addresses. By adding port as a separate config field, we can support ipv6 bootz server addresses.


Chounoki (commented):
Please fix the errors.

gmacf (approved):

Chounoki (changes requested):
Could you also update this file?
https://github.com/openconfig/bootz/blob/main/test/config/bootz_config.textproto

chaskayamane:
Thanks for catching that. test/bootz_config.textproto should also be updated now.

Chounoki (approved):

PR #361, opened 2026-07-16
Enable skipping IDevId serial validation.

chaskayamane:
This change enables skipping IDevId serial validation in the server emulator because some manufacturers may not have IDevId support.

This change also add unit tests for the chassis manager and fixes the client to dial the correct port.

### DIFF
--- a/server/chassismanager/chassismanager_test.go
+++ b/server/chassismanager/chassismanager_test.go
@@ -128,52 +128,33 @@ func TestInMemoryChassisManager(t *testing.T) {
 	})
 
 	// 4. Test GenerateBootstrapData
-
-	serials := []string{serial1, serial2}
-	responses, err := manager.GenerateBootstrapData(ctx, nil, serials)
-	for i, serial := range serials {
-		t.Run(serial, func(t *testing.T) {
-			resp := responses[i]
-			want := &bpb.BootstrapDataResponse{
-				SerialNum:        serial,
-				IntendedImage:    intendedImage,
-				BootPasswordHash: bootPasswordHash,
-				BootConfig:       bootConfig,
-				Pathz:            pathzReq,
-				Authz:            authzReq,
-				CertzProfiles:    certzProfiles,
-				Credentials:      credentials,
-			}
-			if diff := cmp.Diff(want, resp, protocmp.Transform()); diff != "" {
-				t.Errorf("GenerateBootstrapData response mismatch (-want +got):\n%s", diff)
-			}
-		})
-	}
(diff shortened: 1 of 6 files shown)


server/chassismanager/chassismanager_test.go
```diff
@@ -0,0 +1,216 @@
...
+				Authz:            authzReq,
+				CertzProfiles:    certzProfiles,
+				Credentials:      credentials,
+			}
+			if diff := cmp.Diff(want, resp, protocmp.Transform()); diff != "" {
+				t.Errorf("GenerateBootstrapData response mismatch (-want +got):\n%s", diff)
+			}
+		})
+	}
+	t.Run("GenerateBootstrapData Success", func(t *testing.T) {
```
gmacf, line 131: This should probably go above line 132 so that it wraps the entire GenerateBootstrapData section.

server/chassismanager/chassismanager_test.go
```diff
@@ -0,0 +1,216 @@
...
+		err := manager.ResolveChassis(ctx, ch)
+		if err == nil {
+			t.Errorf("ResolveChassis with unknown serial succeeded, want error")
+		}
+	})
+
+	// 4. Test GenerateBootstrapData
+
+	serials := []string{serial1, serial2}
+	responses, err := manager.GenerateBootstrapData(ctx, nil, serials)
```
gmacf: Can you perform the err check here instead of after looking at the responses?

server/chassismanager/chassismanager_test.go
```diff
@@ -0,0 +1,216 @@
...
+				SerialNum:        serial,
+				IntendedImage:    intendedImage,
+				BootPasswordHash: bootPasswordHash,
+				BootConfig:       bootConfig,
+				Pathz:            pathzReq,
+				Authz:            authzReq,
+				CertzProfiles:    certzProfiles,
+				Credentials:      credentials,
+			}
+			if diff := cmp.Diff(want, resp, protocmp.Transform()); diff != "" {
```
gmacf: Nit: you probably don't need to loop over this. You can use cmp.Diff() on the responses slice and compare that with a want slice.

server/chassismanager/chassismanager_test.go
```diff
@@ -0,0 +1,216 @@
...
```
gmacf: Isn't this a duplicate of lines 134 - 147?

chaskayamane: oops it is -- the ai tricked me!

gmacf (approved):

PR #366, opened 2026-08-06
Enforce 10s timeout on Bootz HMAC challenge

gh4683:
Enforce 10s timeout on Bootz HMAC challenge


Chounoki (approved):

Chounoki:
I assume you use a 3rd party library instead of the built-in `time` package only for the test override capability.

Although I guess that can also be achieved by simply sleeping 11 seconds in the test code with the built-in `time` package to get the same test result.

gmacf (approved):

